PR #1, opened 2021-01-20
config.json Updated

PascaliRBX:
I have add %joinedAt% parameter


PascaliRBX:
@SCDerox 
Yes i know but if you want that extra

SCDerox:
It does not really make sense for me. I think it's useless because you can simply check when the message got sent to see when the user joined

PascaliRBX:
Ok reicht mit Englisch ja hast da auch wieder Recht

PR #2, opened 2021-01-20
Updated guildMemberAdd.js

PascaliRBX:
Added %joinedAt%


SCDerox:
This message gets send if a user joins why would we need a %joinedAt% parameter?

PR #6, opened 2021-02-11
Birthday module

SCDerox:
New module!


SCDerox:
Works fine in 5+ guilds; merging

Issue #15, opened 2021-06-11
Config can’t be created on Modules with a ready.js file


jot737:
I wanted to start the bot on my own computer and always when a module with a ready.js file is activated, the bot crashes, bevor the config file can be created, because the config file isn't there.

SCDerox:
Could you please send the following information so I can help you;
* Console output
* Your `module.json`
* Your `ready.js`
* The generation file of your config
Thanks 👍

jot737:
I found the problem and created a new [pull request](https://github.com/SCNetwork/CustomDCBot/pull/16).

SCDerox:
The pull request would break the bot, so I had to close it ^^

SCDerox:
I have pushed some changes regarding generating configuration-files, maybe that helps. Please try updating your bot and if that does not help, please provide me with the required information as described above.

jot737:
Unfortunately, the update did not solve the problem. Here is the required information:
**console output:**
_> customdcbot@1.0.0 start
> node main.js

[MODULE] Module admin-tools is disabled
[MODULE] Module auto-publisher is disabled
[MODULE] Module auto-react is disabled
[MODULE] Loading module betterstatus
[MODULE] Module birthday is disabled
[MODULE] Module giveaways is disabled
[MODULE] Module levels is disabled
[MODULE] Loading module moderation
[DB] Loaded model ./modules/moderation/models/ModerationAction.js
[COMMANDS] Loaded ./modules/moderation/commands/ban.js
[COMMANDS] Loaded ./modules/moderation/commands/clear.js
[COMMANDS] Loaded ./modules/moderation/commands/kick.js
[COMMANDS] Loaded ./modules/moderation/commands/mute.js
[COMMANDS] Loaded ./modules/moderation/commands/removeWarn.js
[COMMANDS] Loaded ./modules/moderation/commands/unban.js
[COMMANDS] Loaded ./modules/moderation/commands/unmute.js
[COMMANDS] Loaded ./modules/moderation/commands/userinfo.js
[COMMANDS] Loaded ./modules/moderation/commands/warn.js
[MODULE] Module partner-list is disabled
[MODULE] Module ping-on-vc-join is disabled
[MODULE] Module reaction-role is disabled
[MODULE] Module serverinfo is disabled
[MODULE] Module suggestions is disabled
[MODULE] Module welcomer is disabled
[COMMANDS] Loaded ./src/commands/help.js
[...]

PR #16, opened 2021-06-11
Fix for issue #15

jot737:
Fix for issue #15 (Config cant be created in modules with ready.js file


SCDerox:
This will not work, as the system is going to check every role and channel while performing the `checkAllConfigs()` function. For that we need to have a client that is actually logged-in to discord.

PR #34, opened 2021-10-03
fix: botReady.js in twitch-notifications

jateute:
fix: why does that worked?


SCDerox:
It's not actually wrong code in a node environment, read more why [here](https://stackoverflow.com/a/26451885), but can lead to bugs as it relies on an object reference.

PR #43, opened 2021-12-25
feat: Added economy-system module

jateute:


modules/economy-system/commands/economy-system.js
```diff
@@ -0,0 +1,378 @@
...
+        if (cooldownModel) {
+            return interaction.reply({
+                content: interaction.str['cooldown'],
+                ephemeral: true
+            });
+        }
+        const cooldown = await model.create({
+            userId: interaction.user.id,
+            command: 'work'
+        });
```
SCDerox, line 26: Duplicated code, try creating a new (private) helper-function which accepts the type of the cooldown as parameter and returns a `Promise<Boolean>` which should describe if the command should get executed further

modules/economy-system/commands/economy-system.js
```diff
@@ -0,0 +1,378 @@
...
+        const moneyToAdd = Math.floor(Math.random() * (interaction.config['maxWorkMoney'] - interaction.config['minWorkMoney'])) + interaction.config['minWorkMoney'];
+        balance(interaction.client, interaction.user.id, 'add', moneyToAdd);
+        interaction.reply({
+            content: interaction.str['workSuccess'].split('%erned%').join(`${moneyToAdd} ${interaction.config['currencySymbol']}`),
+            ephemeral: true
+        });
```
SCDerox, line 32: Please allow embedType, as a lot of our users love creating beautiful embeds ^^
You could use 
```suggestion
   interaction.reply(embedType(interacction.str['workSuccess'], {'%earned%': `${moneyToAdd} ${interaction.config['currencySymbol']}`}, { ephemeral: true });
```
for example

modules/economy-system/commands/economy-system.js
```diff
@@ -0,0 +1,378 @@
...
+                ephemeral: true
+            });
+        }
+        const cooldown = await model.create({
+            userId: interaction.user.id,
+            command: 'daily'
+        });
+        balance(interaction.client, interaction.user.id, 'add', interaction.client.configurations['economy-system']['config']['dailyReward']);
+        interaction.reply({
+            content: `You erned ${interaction.client.configurations['economy-system']['config']['dailyReward']} ${interaction.client.configurations['economy-system']['config']['currencySymbol']} by claiming your daily reward`,
```
SCDerox, line 212: typo ^^
```suggestion
            content: `You earned ${interaction.client.configurations['economy-system']['config']['dailyReward']} ${interaction.client.configurations['economy-system']['config']['currencySymbol']} by claiming your daily reward`,
```

modules/economy-system/commands/economy-system.js
```diff
@@ -0,0 +1,378 @@
...
+            ephemeral: true
+        });
+    }
+};
+
+module.exports.config = {
+    name: 'economy-system',
+    description: 'general economy-system',
+    defaultPermission: true,
+    options: [
```
SCDerox, line 273: I recommend exporting a function here and - based on the configuration - not returning the add, remove and set commands at all. 
You can see an example of this [here](https://github.com/SCNetwork/CustomDCBot/blob/687ac39708ad292c931d2cf3bcb3666ea04076ec/modules/levels/commands/manage-levels.js#L75).

modules/economy-system/configs/strings.json
```diff
@@ -0,0 +1,88 @@
+{
+    "filename": "strings.json",
+    "content": [
+      {
+        "field_name": "notFound",
+        "default": "Cannot find the item",
+        "type": "string",
+        "description": "The message that is send if the item wasn't found",
+        "allowEmbed": false
```
SCDerox, line 9: would be great if embeds were supported everywhere, as they are one of the most loved features (idk why either)

modules/economy-system/economy-system.js
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Add/ Remove xyz from balance/ set balance to
+ * @param {Client} client Client
+ * @param {string} id UserId of the user which is effected
+ * @param {string} action The action which is should be performed (add/ remove/ set)
+ * @param {number} value The value which is added/ removed to/ from the balance/ to which the balance gets set
+ * @returns {Promise<void>}
+ */
+balanceFunction = async function (client, id, action, value) {
```
SCDerox, line 31: try renaming this, maybe to `editBalance`

modules/economy-system/events/ready.js
```diff
@@ -0,0 +1,37 @@
...
+         */
+        function check() {
+            if (client.botReadyAt) return resolve();
+            else {
+                setTimeout(check, 1000);
+            }
+        }
+        setTimeout(check, 10000);
+    });
+}
```
SCDerox, line 23: but... why? Why not use the `botReady` event?

SCDerox (changes requested):
Thank you so much for creating this great module, I especially love features like the the message drops!
I have some (7) suggestions / requested changes - feel free to discuss them if you don't agree with them.
As always, feel free to reach out to me on Discord, if you have any questions. 
Note: I haven't tested your module yet, I just reviewed the code, but will test it as soon as you fixed (some) of my suggestions ^^

modules/economy-system/events/ready.js
jateute: I use the ready event, because otherwise the cooldowns gets reseted every time when the config gets reloaded

SCDerox: Okay... there has to be a better way to do this then looping indefinitely, please don't close this (just) yet, I'm gonna check some stuff an/or implement something more efficient than this ^^

SCDerox: Okay, there are two options to fix this:
1. Add `client.lastReloadAt` paramenter (Date) which gets set at every reload and can be used to determine, if this is the first restart.
2. Instead of using timeouts, store the expiring timestamp in the database and check once a user tries to work if the cooldown passed already. 

I think the second option is the better one, as with this method everyone has to wait for their countdown to expire and reduces database-access (you don't have to delete anything). It also is way more perfroment then the first one, as you don't need to create a timeout on every user request.  

As always, feel free to share your thoughts if you disagree. I am also willing to implement the secound option myself if you currently don't have enough time to invest in this project ^^

SCDerox (commented):
Thanks so much for implementing all my suggestions, I will test your module and approve it asp.

modules/economy-system/commands/economy-system.js
```diff
@@ -0,0 +1,298 @@
...
+        const moneyToAdd = Math.floor(Math.random() * (interaction.config['maxWorkMoney'] - interaction.config['minWorkMoney'])) + interaction.config['minWorkMoney'];
+        balance(interaction.client, interaction.user.id, 'add', moneyToAdd);
+        interaction.reply(embedType(interaction.str['workSuccess'], {'%earned%': `${moneyToAdd} ${interaction.config['currencySymbol']}`}, { ephemeral: true }));
+        createleaderboard(interaction.client);
+        interaction.client.logger.info(`[economy-system] The user ${interaction.user.username}#${interaction.user.discriminator} gained ${moneyToAdd} ${interaction.config['currencySymbol']} by working`);
+        if (interaction.client.logChannel) interaction.client.logChannel.send(`[economy-system] The user ${interaction.user.username}#${interaction.user.discriminator} gained ${moneyToAdd} ${interaction.config['currencySymbol']} by working`);
+    },
+    'crime': async function (interaction) {
+        if (!cooldown('work', interaction.config['crimeCooldown'] * 60000, interaction.user.id)) return interaction.reply(embedType(interaction.str['cooldown'], {}, { ephemeral: true }));
```
SCDerox, line 53: Wrong cooldown type here (and also in the following subcommands)
```suggestion
        if (!cooldown('crime', interaction.config['crimeCooldown'] * 60000, interaction.user.id)) return interaction.reply(embedType(interaction.str['cooldown'], {}, { ephemeral: true }));
```

SCDerox (changes requested):
Great work so far, I ran into some issues when trying to test it, so I documented them here. It would be great if you could fix these, after that I will be able to continue my review ^^

modules/economy-system/commands/economy-system.js
```diff
@@ -0,0 +1,282 @@
...
+                id: user.id
+            }
+        });
+        if (!balanceV) return interaction.reply(embedType(interaction.str['userNotFound']), {'%user%': `${interaction.user.username}#${interaction.user.discriminator}`});
+        interaction.reply(embedType(interaction.str['balanceReply'], {'%user%': `${user.username}#${user.discriminator}`, '%balance%': `${balanceV['dataValues']['balance']} ${interaction.client.configurations['economy-system']['config']['currencySymbol']}`}, { ephemeral: true }));
+    }
+};
+
+module.exports.config = {
+    name: 'economy-system',
```
SCDerox, line 171: i would suggest renaming the command to `/economy`, as it is shorter and has basically the same meaning

modules/economy-system/events/messageCreate.js
```diff
@@ -0,0 +1,17 @@
...
+    if (!client.botReadyAt) return;
+    if (!message.guild) return;
+    if (message.author.bot) return;
+    if (message.guild.id !== client.config.guildID) return;
+
+    const config = client.configurations['economy-system']['config'];
+
+    if (config['messageDrops'] === 0) return;
+    if (Math.floor(Math.random() * config['messageDrops']) !== 1) return;
+    const toAdd = Math.floor(Math.random() * (interaction.config['messageDropsMax'] - interaction.config['messageDropsMin'])) + interaction.config['messageDropsMin'];
```
SCDerox, line 13:
```suggestion
    const toAdd = Math.floor(Math.random() * (config['messageDropsMax'] - config['messageDropsMin'])) + config['messageDropsMin'];
```
interaction is not defined here

modules/economy-system/events/messageCreate.js
```diff
@@ -0,0 +1,17 @@
...
+    balance(client, message.author.id, 'add', toAdd);
```
SCDerox, line 14:
```suggestion
    await balance(client, message.author.id, 'add', toAdd);
```
balance() returns a Promise

modules/economy-system/commands/shop.js
```diff
@@ -0,0 +1,133 @@
...
+        interaction.reply(embedType(interaction.client.configurations['economy-system']['strings']['buyMsg'], {'%item%': itemName['value']}, { ephemeral: true }));
+        interaction.client.logger.info(`[economy-system] The user ${interaction.user.username}#${interaction.user.discriminator} has buyed the shop item ${itemName['value']}`);
+        if (interaction.client.logChannel) interaction.client.logChannel.send(`[economy-system] The user ${interaction.user.username}#${interaction.user.discriminator} has buyed the shop item ${itemName['value']}`);
+    },
+    'list': async function (interaction) {
+        const msg = await createShopMsg(interaction.client);
+        interaction.reply({
+            content: msg,
+            ephemeral: true
+        });
```
SCDerox, line 54: commands does not work ):
Could you check if it works for you (tries to send empty message)

modules/economy-system/commands/economy-system.js
```diff
@@ -0,0 +1,282 @@
...
+        if (interaction.client.logChannel) interaction.client.logChannel.send(`[economy-system] The user ${interaction.user.username}#${interaction.user.discriminator} gained ${moneyToAdd} ${interaction.config['currencySymbol']} by doing crime`);
+    },
+    'rob': async function (interaction) {
+        if (!await cooldown('rob', interaction.config['robCooldown'] * 60000, interaction.user.id, interaction.client)) return interaction.reply(embedType(interaction.str['cooldown'], {}, { ephemeral: true }));
+        const user = await interaction.options.getUser('user');
+        const robbedUser = await interaction.client.models['economy-system']['Balance'].findOne({
+            where: {
+                id: user.id
+            }
+        });
```
SCDerox, line 69: please handle if robbedUser is null (user has no economy profile), causes crash currently

modules/economy-system/commands/economy-system.js
```diff
@@ -0,0 +1,282 @@
...
+    },
+    'balance': async function (interaction) {
+        let user = interaction.options.getUser('user');
+        if (!user) user = interaction.user;
+        const balanceV = await interaction.client.models['economy-system']['Balance'].findOne({
+            where: {
+                id: user.id
+            }
+        });
+        if (!balanceV) return interaction.reply(embedType(interaction.str['userNotFound']), {'%user%': `${interaction.user.username}#${interaction.user.discriminator}`});
```
SCDerox, line 165:
```suggestion
        if (!balanceV) return interaction.reply(embedType(interaction.str['userNotFound']), {'%user%': `${interaction.user.username}#${interaction.user.discriminator}`, { ephemeral: true }});
```

SCDerox (changes requested):
looking good so far, I found some small issues, sorry for the delay

modules/economy-system/commands/economy-system.js
```diff
@@ -60,13 +60,14 @@ module.exports.subcommands = {
...
     },
     'rob': async function (interaction) {
-        if (!await cooldown('rob', interaction.config['robCooldown'] * 60000, interaction.user.id, interaction.client)) return interaction.reply(embedType(interaction.str['cooldown'], {}, { ephemeral: true }));
         const user = await interaction.options.getUser('user');
         const robbedUser = await interaction.client.models['economy-system']['Balance'].findOne({
             where: {
                 id: user.id
             }
         });
+        if (!robbedUser) return interaction.reply(embedType(interaction.str['userNotFound']), {'%user%': `${interaction.user.username}#${interaction.user.discriminator}`}, { ephemeral: true });
```
SCDerox:
```suggestion
        if (!robbedUser) return interaction.reply(embedType(interaction.str['userNotFound'], {'%user%': `${interaction.user.username}#${interaction.user.discriminator}`}, { ephemeral: true }));
```

modules/economy-system/commands/economy-system.js
```diff
@@ -162,13 +163,13 @@ module.exports.subcommands = {
                 id: user.id
             }
         });
-        if (!balanceV) return interaction.reply(embedType(interaction.str['userNotFound']), {'%user%': `${interaction.user.username}#${interaction.user.discriminator}`});
+        if (!balanceV) return interaction.reply(embedType(interaction.str['userNotFound']), {'%user%': `${interaction.user.username}#${interaction.user.discriminator}`}, { ephemeral: true });
```
SCDerox, line 166: should be part of embedType
```suggestion
        if (!balanceV) return interaction.reply(embedType(interaction.str['userNotFound'], {'%user%': `${interaction.user.username}#${interaction.user.discriminator}`}, { ephemeral: true }));
```

SCDerox (changes requested):
two very minor things, I'm gonna fix them myself and merge then, thank you so much for your contribution <3

SCDerox (approved):
Perfekt, again, thank you so much for your contribution!

PR #45, opened 2022-01-16
feat: Added Userfeedback to the economy-system

jateute:


default-locales.json
```diff
@@ -26,7 +26,9 @@
       "loaded-cli": "Loaded API-Action %c in %p",
       "channel-lock": "Locked channel",
       "channel-unlock": "Unlocked channel",
-      "channel-unlock-data-not-found": "Unlocking channel with ID %c failed because it was never locked (which is weird to beginn with)."
+      "channel-unlock-data-not-found": "Unlocking channel with ID %c failed because it was never locked (which is weird to beginn with).",
+      "migrate-start": "Starting migrating the model %o by the module %m. Please don't stop the Bot",
```
SCDerox, line 30:
```suggestion
      "migrate-start": "Starting migration of the model %o to the model %m. Please don't stop the Bot",
```

modules/economy-system/cli.js
```diff
@@ -44,7 +44,7 @@ module.exports.commands = [
         command: 'balance',
         description: 'Show all balances from the DataBase',
         run: async function (input) {
-            const balances = await input.client.models['economy-system']['Balance'].findAll();
+            const balances = await input.client.models['economy-system']['NewBalance'].findAll();
```
SCDerox, line 47: I would suggest naming the old model `BalanceOld` (or similar) and leave the name of the current model at `Balance`. 

(this is just a suggestion, you can leave it as it is, if you want)

modules/economy-system/configs/strings.json
```diff
@@ -7,20 +7,30 @@
   "content": [
     {
       "field_name": "notFound",
-      "default": "This item could not be found",
+      "humanname-en": "not found",
+      "humanname-de": "Nicht Gefunden",
```
SCDerox, line 11: I would suggest calling it something like "Not-Found-Message" / "Nicht-Gefunden-Nachricht" as it is a message

SCDerox (commented):
I have some small suggestions, but nothing big, thanks so much for your contribution!
I didn't have the time to test it yet (I have a latin lesson while typing this), but will do it as soon as possible.

src/functions/helpers.js
```diff
@@ -340,4 +339,28 @@ async function unlockChannel(channel, reason = localize('main', 'channel-unlock'
...
+async function migrate(module, oldModel, newModel, client) {
+    const old = await client.models[module][oldModel].findAll();
+    if (old.length === 0) return;
+    client.logger.info(localize('main', 'migrate-start', {o: oldModel, m: newModel}));
+    await old.forEach(async (model) => {
+        delete model.dataValues.updatedAt;
+        delete model.dataValues.createdAt;
+        await client.models[module][newModel].create(model.dataValues);
+        await model.destroy();
+    });
```
SCDerox, line 362: Promise returned from forEach argument ist ignored
```suggestion
    for (const model of old) {
        delete model.dataValues.updatedAt;
        delete model.dataValues.createdAt;
        await client.models[module][newModel].create(model.dataValues);
        await model.destroy();
    }
```

modules/economy-system/events/messageCreate.js
```diff
@@ -10,6 +10,7 @@ module.exports.run = async function (client, message) {
     const config = client.configurations['economy-system']['config'];
 
     if (config['messageDrops'] === 0) return;
+    if (config['msgDropsIgnoredChannels'].includes(message.channel.id)) return;
     if (Math.floor(Math.random() * config['messageDrops']) !== 1) return;
     const toAdd = Math.floor(Math.random() * (config['messageDropsMax'] - config['messageDropsMin'])) + config['messageDropsMin'];
     await balance(client, message.author.id, 'add', toAdd);
```
SCDerox, line 16: balance is not a function, use `editBalance` instead
```suggestion
    await editBalance(client, message.author.id, 'add', toAdd);
```

SCDerox (changes requested):
two small things, fixed them internally, will be merging as sonn as you have fixed them ^^

PR #50, opened 2022-02-06
Fixes and updates

jateute:
Fix: Fixed errors in the econmy-system
Added Userfeedback to the economy-system


SCDerox:
Thanks so much! Does this depended on the new v3 version or is this still for V2?

jateute:
> Thanks so much! Does this depended on the new v3 version or is this still for V2?

It depends on V2, because I didn't know whether v3 is already ready.

SCDerox:
Okay, perfect, we are still working on V3 and it is still in the development phase. We will roll-out V3 with the new Dashboard, so we can publish your changes without needing to wait for that. 

This pull-request includes *some* commits that are already in this branch... Could you try to minimize the amount of commits somehow? For example by forking the main-branch another time and only commiting the new changes?

PR #57, opened 2022-05-01
Added customizable strings.json for command answers.

hfgd123:
As requested by Shini I've now added a config file to customize the answers to the commands.


hfgd123:
ok. I have no idea what exactly went wrong here but it looks like it tries to put every single commit I did before the last merge into this pull request as well

SCDerox (approved):
Thanks for your contribution! 
Please do not include all already merged commit-messages in PRs in the future.

PR #59, opened 2022-05-03
Created module status-roles

hfgd123:
I have created another module based on an idea by Shini

### DIFF
--- a/modules/status-roles/events/presenceUpdate.js
+++ b/modules/status-roles/events/presenceUpdate.js
@@ -1,4 +1,4 @@
-const {localize} = require("../../../src/functions/localize");
+const {localize} = require('../../../src/functions/localize');
 
 module.exports.run = async function (client, newPresence) {
 
@@ -19,4 +19,4 @@ module.exports.run = async function (client, newPresence) {
             }
         }
     }
-}
+};
--- a/modules/status-roles/events/presenceUpdate.js
+++ b/modules/status-roles/events/presenceUpdate.js
@@ -22,6 +22,10 @@ module.exports.run = async function (client, oldPresence, newPresence) {
     } else {
         removeRoles();
     }
+
+    /**
+     * Removes the roles of an user who no longer fulfills the criteria
+     */
     function removeRoles() {
         for (let i = 0; i < roles.length; i++) {
             if (member.roles.cache.has(roles[i])) {
--- a/modules/status-roles/events/presenceUpdate.js
+++ b/modules/status-roles/events/presenceUpdate.js
@@ -1,20 +1,29 @@
 const {localize} = require('../../../src/functions/localize');
 
-module.exports.run = async function (client, newPresence) {
+module.exports.run = async function (client, oldPresence, newPresence) {
 
     if (!client.botReadyAt) return;
+    if (newPresence.member.guild.id !== client.guildID) return;
     const moduleConfig = client.configurations['status-roles']['config'];
     const roles = moduleConfig.roles;
     const status = moduleConfig.words;
     const member = newPresence.member;
 
     if (newPresence.activities.length > 0) {
-        if (status.some(word => newPresence.activities[0].state.includes(word))) {
-            return member.roles.add(roles, localize('mass-role', 'fulfilled'));
-        } else {
+        if (newPresence.activities[0].state) {
+            if (status.some(word => newPresence.activities[0].state.toLowerCase().includes(word.toLowerCase()))) {
+                return member.roles.add(roles, localize('status-role', 'fulfilled'));
+            } else {
+                for (let i = 0; i < roles.length; i++) {
+                    if (member.roles.cache.has(roles[i])) {
+                        member.roles.remove(roles[i], localize('status-role', 'not-fulfilled'));
+                    }
+                }
+            }
+        }  else {
             for (let i = 0; i < roles.length; i++) {
                 if (member.roles.cache.has(roles[i])) {
-                    member.roles.remove(roles[i], localize('mass-role', 'not-fulfilled'));
+                    member.roles.remove(roles[i], localize('status-role', 'not-fulfilled'));
                 }
             }
         }
--- a/modules/status-roles/events/presenceUpdate.js
+++ b/modules/status-roles/events/presenceUpdate.js
@@ -14,17 +14,18 @@ module.exports.run = async function (client, oldPresence, newPresence) {
             if (status.some(word => newPresence.activities[0].state.toLowerCase().includes(word.toLowerCase()))) {
                 return member.roles.add(roles, localize('status-role', 'fulfilled'));
             } else {
-                for (const key in roles) {
-                    if (member.roles.cache.has(roles[key])) {
-                        member.roles.remove(roles[key], localize('status-role', 'not-fulfilled'));
-                    }
-                }
+                removeRoles();
             }
         } else {
-            for (let i = 0; i < roles.length; i++) {
-                if (member.roles.cache.has(roles[i])) {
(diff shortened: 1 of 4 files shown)


modules/status-roles/events/presenceUpdate.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports.run = async function (client, newPresence) {
+
+    if (!client.botReadyAt) return;
+    const moduleConfig = client.configurations['status-roles']['config'];
+    const roles = moduleConfig.roles;
+    const status = moduleConfig.words;
+    const member = newPresence.member;
+
+    if (newPresence.activities.length > 0) {
+        if (status.some(word => newPresence.activities[0].state.includes(word))) {
```
SCDerox, line 12: - `newPresence.activities[0].state` can be null, as definied [here](https://discord.js.org/#/docs/discord.js/stable/class/Activity?scrollTo=state)
- This is case sensitive, but that's not ideal: Try lower (or upper)-casing both values

modules/status-roles/events/presenceUpdate.js
```diff
@@ -0,0 +1,22 @@
+const {localize} = require("../../../src/functions/localize");
+
+module.exports.run = async function (client, newPresence) {
+
+    if (!client.botReadyAt) return;
+    const moduleConfig = client.configurations['status-roles']['config'];
```
SCDerox, line 6: Please check if the member is on the guild the bot runs on first. 
Example: `if (newPresence.member.guild.id !== client.guildID) return`

modules/status-roles/events/presenceUpdate.js
```diff
@@ -0,0 +1,22 @@
+const {localize} = require("../../../src/functions/localize");
+
+module.exports.run = async function (client, newPresence) {
```
SCDerox, line 3:
```suggestion
module.exports.run = async function (client, oldPresence, newPresence) {
```
I belive you use the oldPresence object currently, instead of the new presence object, please consult the [discord.js docs](https://discord.js.org/#/docs/discord.js/stable/class/Client?scrollTo=e-presenceUpdate)

modules/status-roles/events/presenceUpdate.js
```diff
@@ -0,0 +1,22 @@
...
+    const moduleConfig = client.configurations['status-roles']['config'];
+    const roles = moduleConfig.roles;
+    const status = moduleConfig.words;
+    const member = newPresence.member;
+
+    if (newPresence.activities.length > 0) {
+        if (status.some(word => newPresence.activities[0].state.includes(word))) {
+            return member.roles.add(roles, localize('mass-role', 'fulfilled'));
+        } else {
+            for (let i = 0; i < roles.length; i++) {
```
SCDerox, line 15: Note: You could simplify this by using

- `for (const role of roles)`
- `for (const key in roles)

(better readability)

SCDerox (changes requested):
Great module, thank you!
Here are some things I noticied when reading though the code

modules/status-roles/events/presenceUpdate.js
```diff
@@ -0,0 +1,22 @@
...
+
+    if (!client.botReadyAt) return;
+    const moduleConfig = client.configurations['status-roles']['config'];
+    const roles = moduleConfig.roles;
+    const status = moduleConfig.words;
+    const member = newPresence.member;
+
+    if (newPresence.activities.length > 0) {
+        if (status.some(word => newPresence.activities[0].state.includes(word))) {
+            return member.roles.add(roles, localize('mass-role', 'fulfilled'));
```
SCDerox, line 13:
```suggestion
            return member.roles.add(roles, localize('status-role', 'fulfilled'));
```

modules/status-roles/events/presenceUpdate.js
```diff
@@ -0,0 +1,22 @@
...
+        } else {
+            for (let i = 0; i < roles.length; i++) {
+                if (member.roles.cache.has(roles[i])) {
+                    member.roles.remove(roles[i], localize('mass-role', 'not-fulfilled'));
```
SCDerox, line 17:
```suggestion
                    member.roles.remove(roles[i], localize('status-role', 'not-fulfilled'));
```

modules/status-roles/events/presenceUpdate.js
hfgd123: Where am I using oldPresence? 🤔 Afaik what I wrote just depends on what is in the Presence after the change

SCDerox: No, the order is relevant, not the name. As describe in the documentation, `oldPresence` is passed first, while ´newPresence` is the secound parameter (here third because we add `client`).
With the current code, users may have the role even when they have changed their status.

hfgd123: oh. that makes sense. my bad

SCDerox: image
image

modules/status-roles/events/presenceUpdate.js
hfgd123: ok. i understand, that this is posible. but how do I get the i for checking and removing all roles then? If I just write the i++ into the loop it would be almost the same as it is now

SCDerox: `roles` is an array with objects inside them. You can get an object at a position by using `roles[i]` (this is what you do now). 

- With `for (const role of roles)` you get the object without needing to query it yourself. (`role` would be equal to `roles[i]`)
- WIth `for (const key in roles)` you loop though each key (corosponding to position in arrays). In your case, the `i` would be `key`. (You could also just write `for (const i in roles)`

Note that this was only a note for better readability (dx) and can be ignored, if you don't want to include this change. Press "Resolve conversation" below to do so.

hfgd123: I already included it. It is always interesting to learn new stuff (as long as it isn't latin)

SCDerox: i agree.

modules/status-roles/events/presenceUpdate.js
```diff
@@ -0,0 +1,31 @@
...
+module.exports.run = async function (client, oldPresence, newPresence) {
+
+    if (!client.botReadyAt) return;
+    if (newPresence.member.guild.id !== client.guildID) return;
+    const moduleConfig = client.configurations['status-roles']['config'];
+    const roles = moduleConfig.roles;
+    const status = moduleConfig.words;
+    const member = newPresence.member;
+
+    if (newPresence.activities.length > 0) {
```
SCDerox, line 12: If this is false, the roles need to be removed, as there is no status

modules/status-roles/events/presenceUpdate.js
```diff
@@ -0,0 +1,31 @@
...
+        if (newPresence.activities[0].state) {
+            if (status.some(word => newPresence.activities[0].state.toLowerCase().includes(word.toLowerCase()))) {
+                return member.roles.add(roles, localize('status-role', 'fulfilled'));
+            } else {
+                for (const key in roles) {
+                    if (member.roles.cache.has(roles[key])) {
+                        member.roles.remove(roles[key], localize('status-role', 'not-fulfilled'));
+                    }
+                }
```
SCDerox, line 21: Duplicated code (24-26). Try to build a helper-function to increase maintainability.

SCDerox (changes requested):
to other small things ^^

SCDerox (approved):
Thanks for your contribution!

The version for SCNX got released ^^

PR #62, opened 2022-05-19
Created Color-Me module

hfgd123:
Created Color-Me module. Requesting review to check for mergablilty

### DIFF
--- a/default-locales.json
+++ b/default-locales.json
@@ -400,6 +400,18 @@
     "status-role": {
       "fulfilled": "Status-role condition is fulfilled",
       "not-fulfilled": "Status-role condition is no longer fulfilled"
+    },
+    "color-me": {
+      "create-log-reason": "%user redeemed their boosting-rewards by requesting the creation of this role",
+      "edit-log-reason": "%user edited their boosting-reward-role",
+      "delete-unboost-log-reason": "%user stopped boosting, so their role got deleted",
+      "delete-manual-log-reason": "%user deleted their role manually",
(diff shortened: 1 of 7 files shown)


hfgd123:
Requesting review by @SCDerox

SCDerox (changes requested):
Looking pretty good! 
While I haven't done my usual code-quality check (if you want me to do that, please message me), there are a few issues I encountered when using this module:

- Position of the new role is always at the bottom. Staff should decide below which role the new role should be created (message me if you need help with that)
- Show the user how long the need to wait until their cooldown expires (use `dateToDiscordTimestamp(role.timestamp, 'R'` from helpers.js)
- Entering something random into the color field will show an error about the max-role count, which is not optimal as the input is invalid rather then the role-count is maxed-out

SCDerox:
Also sorry for the rate review, was quite busy working on something :sweat:

hfgd123:
> Show the user how long the need to wait until their cooldown expires (use dateToDiscordTimestamp(role.timestamp, 'R' from helpers.js)

wouldn't this just tell the user, when they edited their role the last time? How can I do math things (english 10/10 :hAA:) with dates?

SCDerox:
Oh, yeah, I am brain dead already (imagine sleeping at night); I will get back to you in about an hour, I have to do some shit irl

SCDerox:
I believe `dateToDiscordTimestamp(new Date(role.timestamp.getTime() + moduleConf['updateCooldown'] * 3600000), 'R')` should work

SCDerox (approved):
Everything works for me, thanks for your contribution!

PR #64, opened 2022-05-29
Welcomer

hfgd123:
Added option in the welcomer module to delete the welcome messages of users when they leave within 7 days after their join

### DIFF
--- a/modules/auto-react/config.json
+++ b/modules/auto-react/config.json
@@ -22,21 +22,8 @@
       "humanname-de": "Erwähnungen",
       "humanname-en": "Mentions",
       "type": "keyed",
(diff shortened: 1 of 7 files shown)


SCDerox (dismissed):
It works - but trying to fetch the message in each channel is kinda inefficient, I would suggest storing the channel-id in the database as well

SCDerox:
If you want, I can merge it as it is, or you could implement the suggested change - in either case, please let me know

hfgd123:
I'll try to implement your suggestion

SCDerox:
FYI: I will push some close-sourced changes we added to the welcomer module, should not interfere with your changes ^^

SCDerox (approved):
lgtm, thanks for your contribution!

PR #66, opened 2022-06-01
Committing SCDerox's auto-react update, after understanding it

hfgd123:
(at least I think so)


SCDerox:
totally forgot that this exists sorry

PR #79, opened 2022-10-10
feat(economy-system): Added ability to make command-replies public

jateute:


modules/economy-system/configs/config.json
```diff
@@ -196,6 +196,16 @@
       "default": 100,
       "type": "integer",
       "description": "The weekly reward"
+    },
+    {
+      "field_name": "PublicCommandReplies",
```
SCDerox, line 201: We usually try to keep config field_names starting with a lower letter, I suggest renaming it if possible (the whole config files has lowercase-field-names except this option)

SCDerox (changes requested):
Looking pretty good, found one small issue, could you please check that?

PR #107, opened 2023-02-20
Add Connect Four module

DEVTomatoCake:
(Also fixes an oversight in the rock-paper-scissors module & changes the indention in en.json to match the one in de.json)


SCDerox (approved):
Thank you for your contribution, your module will be available on SCNX shortly.

PR #110, opened 2023-02-25
Add uno module

DEVTomatoCake:
Adds an uno module (pain)
Also changes/fixes:
- `[object Object]` if a slash command from a disabled module is used
- Users being able to use bots and themselves in Connect Four
- Highlights the part which lead to a win in Connect Four


DEVTomatoCake:
Cannot reproduce.

SCDerox (approved):
Sorry for the delays on this review. We're always working hard to respond to pull requests as fast as possible, but in this case personal reasons (school lol) lead to delays.
Nevertheless, thank you for your contribution :heart: 
Your changes should be available on SCNX shortly.

PR #127, opened 2023-09-24
Add starboard feature

DEVTomatoCake:
Resolves https://features.sc-network.net/board/2/597

Also adds a feature to sticky messages which resends them if the user changed it's configuration.


modules/sticky-messages/events/messageCreate.js
```diff
@@ -3,37 +3,38 @@ const channelData = {};
...
+ * @param {Snowflake} clientId User ID of the bot
+ * @param {Discord.TextBasedChannel} channel
  */
-async function deleteMessage(client, msg) {
+async function deleteMessage(clientId, channel) {
     let message;
-    message = await msg.channel.messages.fetch(channelData[msg.channel.id].msg).catch(async () => {
-        const msgs = await msg.channel.messages.fetch({limit: 20});
-        message = msgs.find(m => m.author.id === client.user.id);
+    message = await channel.messages.fetch(channelData[channel.id].msg).catch(async () => {
```
SCDerox, line 11: `channelData[channel.id]` might be `undefined` if the module configuration gets changed and reloaded before messages where sent in the configured channel, leading to a TypeError

modules/starboard/handleStarboard.js
```diff
@@ -0,0 +1,63 @@
...
+module.exports = async (client, msgReaction, isReactionRemove = false) => {
+    if (!client.botReadyAt) return;
+    const msg = msgReaction.message;
+    if (!msg.guild) return;
+    if (msg.guild.id !== client.guildID) return;
+    if (msgReaction.partial) msgReaction = await msgReaction.fetch();
+
+    const starConfig = client.configurations['starboard']['config'];
+    if (!starConfig || (starConfig.emoji !== msgReaction.emoji.name && starConfig.emoji !== msgReaction.emoji.id)) return;
+    if (isNaN(starConfig.minStars)) return disableModule('starboard', localize('starboard', 'invalid-minstars', {stars: starConfig.minStars}));
```
SCDerox, line 12: `localize` is not imported in this file. Please add an import statement

SCDerox: Neither is disableModule, as a matter of fact

modules/starboard/handleStarboard.js
```diff
@@ -0,0 +1,63 @@
...
+        if (isReactionRemove) {
+            if (starboardMsg) starboardMsg.delete();
+            client.models['starboard']['StarMsg'].destroy({
+                where: {
+                    msgId: msg.id
+                }
+            });
+        }
+        return;
+    }
```
SCDerox, line 38: While this is pretty cool, please add some kind of mechanism to protect against users spamming the star reaction. I would suggest that a user can only star up to 5 messages per hour or something like that (this also makes stars more valuable, maybe make such a limit configurable?)

SCDerox (changes requested):
Hi - thanks so much for your suggestion! I've found a few issues, please review them ^^

modules/starboard/handleStarboard.js
DEVTomatoCake: ESLint should be configured to flag that

SCDerox: good point, feel fee to add that or I'll look into that

modules/starboard/handleStarboard.js
DEVTomatoCake: Hm I can def implement something against spamming reactions but those wouldn't be stored in a database.

I do get the idea, however this is difficult to implement with reactions - how would the user know how long they have to wait until they can react again & if their reaction counted? I think this would lead to even more reaction spam (esp. if the bot removes invalid reactions).
I'd rather implement a delay for editing the starboard message so they bot is unlikely to get ratelimited there, and hope that server admins are able to configure the module accordingly to their servers activity.
I'll just ask on the original feature board page again.

SCDerox: I would suggest removing the reactions when users are out of their configured star-amount and message them. I agree that a message-spam-limit would be useful. This is not about Discord ratelimiting the bot, instead about UX on the servers (imagine having to mark a channel as read every minute because someone is spamming star on random messages).

DEVTomatoCake: That's why there's the minimum star amount, this should never be set to 1.

SCDerox: Sure, but there will be situations where only one star is missing. A user then can spam the star to generate messages in the starboard-channel. 
Additionally, the community might abuse this feature to star every message (because it's fun you know), that's why I think a star-limit per user per timeframe would be a good idea.

modules/starboard/handleStarboard.js
```diff
@@ -0,0 +1,64 @@
...
+
+    const generatedMsg = await embedTypeV2(starConfig.message, {
+        '%stars%': msgReaction.count,
+        '%content%': msg.content,
+        '%username%': msg.author.username,
+        '%usertag%': formatDiscordUserName(msg.author),
+        '%nickname%': msg.member ? msg.member.displayName : msg.author.username,
```
SCDerox, line 46: In contrast to `.nickname`, `.displayName` will automatically default to msg.author.username[^1].

[^1]: https://github.com/discordjs/discord.js/blob/759c0b0d46ce67f4208cdf22214d2272860f27b9/src/structures/GuildMember.js#L267

modules/starboard/configs/config.json
```diff
@@ -0,0 +1,209 @@
...
+						"de": "Der Nickname des Autors"
+					}
+				},{
+					"name": "userTag",
+					"description": {
+						"en": "The tag of the author of the starred message",
+						"de": "Der Tag des Autors der gestarrten Nachricht"
+					}
+				},{
+					"name": "userAvatar",
```
SCDerox, line 105:
```suggestion
					"name": "userAvatar",
					 "isImage": true,
```
`isImage` is needed to allow users to use this parameter in the dashboard as an image

modules/starboard/configs/config.json
```diff
@@ -0,0 +1,209 @@
...
+						"de": "Die Kanalerwähnung des Kanals, in dem die gestarrte Nachricht gesendet wurde"
+					}
+				},{
+					"name": "emoji",
+					"description": {
+						"en": "The set starboard emoji for lazy users",
+						"de": "Das festgelegte Starboard-Emoji für faule Nutzer"
+					}
+				},{
+					"name": "image",
```
SCDerox, line 129:
```suggestion
					"name": "image",
				       "isImage": true,
```
`isImage` is needed to allow users to use this parameter in the dashboard as an image

SCDerox (changes requested):
Hi, I found this very small issue. Additionally I observed that when a user has used all their stars in an hour removes their reaction from a starred message. this doesn't trigger an updated. This means that the star count is out-of-sync in the starred message and in the starboard message. This can create confusion (also this might lead to messages with less then the required star amount in the starboard). Please let me know if this intended behavior.

DEVTomatoCake:
> Additionally I observed that when a user has used all their stars in an hour removes their reaction from a starred message. this doesn't trigger an updated. This means that the star count is out-of-sync in the starred message and in the starboard message. This can create confusion (also this might lead to messages with less then the required star amount in the starboard). Please let me know if this intended behavior.

The original intention was to prevent spamming the reaction, so yes, it's intended - if you think spamming won't cause trouble, I'll change it.

SCDerox:
While I agree, this is a suboptimal solution from my perspective: Instead of limiting the ability to remove the reaction (or rather count the removal of reaction), we should prevent users from adding new reactions. The reason for this is simple: We can't prevent a reaction from being removed, but we an prevent (or undo) a emoji being added. This ensures that the amount of stars under the message stays accurate with the amount of stars displayed in the starboard - from the users perspective, the current behavior appears as a bug, while preventing them from adding a new emoji is easier to understand.

DEVTomatoCake:
The preventing thing already happens, doesn't it?

SCDerox:
Yes, this is just about how a reaction removed after a user got "ratelimited" does not count as a removed reaction

modules/starboard/handleStarboard.js
```diff
@@ -0,0 +1,99 @@
...
+        where: {
+            userId: user.id,
+            createdAt: {
+                [Op.gt]: Date.now() - 1000 * 60 * 60
+            }
+        }
+    });
+
+    if (!isReactionRemove) {
+        if (starUser.length >= starConfig.starsPerHour) {
```
SCDerox, line 31:
```suggestion
        if (!isReactionRemove && starUser.length >= starConfig.starsPerHour) {
```
just make it one :wink:

SCDerox (approved):
Sorry about that - everything looks good to go :+1:

SCDerox:
Thanks for your contribution - you have received a payout as part of our [Open Source Developer Program](https://faq.scnx.app/open-source-developer-pool/). You can find details in your SCNX organisation payout history. To pay out the received amount, please follow instructions in your dashboard. If you have any questions, please contact ScootKit staff on [scnx.app/help](https://scnx.app/help).
Cheers ✌️

PR #128, opened 2023-10-07
Fixed a 2 things in the economy-system

jateute:


modules/economy-system/economy-system.js
```diff
@@ -233,8 +233,8 @@ async function leaderboard(client) {
         .setFooter({text: client.strings.footer, iconURL: client.strings.footerImgUrl});
 
     if (model.length !== 0) embed.addField('Leaderboard:', await topTen(model, client));
-    if (moduleStr['leaderboardEmbed']['thumbnail']) embed.setThumbnail(moduleStr['leaderboardEmbed']['thumbnail']);
-    if (moduleStr['leaderboardEmbed']['image']) embed.setImage(moduleStr['leaderboardEmbed']['image']);
+    if (moduleStr['leaderboardEmbed']['thumbnail'] !== ' ') embed.setThumbnail(moduleStr['leaderboardEmbed']['thumbnail']);
+    if (moduleStr['leaderboardEmbed']['image'] !== ' ') embed.setImage(moduleStr['leaderboardEmbed']['image']);
```
DEVTomatoCake, line 237:
```suggestion
    if (moduleStr['leaderboardEmbed']['thumbnail'].trim() !== '') embed.setThumbnail(moduleStr['leaderboardEmbed']['thumbnail']);
    if (moduleStr['leaderboardEmbed']['image'].trim() !== '') embed.setImage(moduleStr['leaderboardEmbed']['image']);
```

Comparing against one space doesn't seem to make sense, this makes it even work if users use two spaces! :hAA:

jateute: I did it, because one space is for some reason the value it has when you don’t add :hAA: But fair point ^^

SCDerox (approved):
Thanks for fixing these issues!

PR #130, opened 2023-10-08
Sticky messages setting to allow responses to bots

DEVTomatoCake:
Small change to allow enabling responding to other bots per sticky message.
The current bot still gets ignored to prevent reacting to itself.


SCDerox (approved):
Thanks for fixing the issues I mentioned on Discord!

PR #142, opened 2024-06-17
added nickname module

hfgd123:
Added overdue nickname module as per Hörnchen's suggestion


modules/nickname/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,67 @@
...
+    const roles = client.configurations['nickname']['config'];
+    const moduleModel = client.models['nickname']['User'];
+
+    let rolePrefix = "";
+    let hoistrole
+    if (newGuildMember.roles.hoist === null) {
+        hoistrole = null;
+    } else {
+        hoistrole = newGuildMember.roles.hoist.id;
+    }
```
SCDerox, line 16:
```suggestion
    let hoistrole; // null by default, you can also use "hoistrole = null;" if you want.
    if (newGuildMember.roles.hoist) hoistrole = newGuildMember.roles.hoist.id;
```

modules/nickname/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,67 @@
...
+
+    for (const role of roles) {
+        if (role.roleID === hoistrole) {
+            rolePrefix = role.prefix;
+        }
+    }
```
SCDerox, line 22: why is this a loop?

```suggestion
   rolePrefix = roles.find(r => r.roleID === hoistrole)?.prefix; 
```

The `?` ensures that a possible null value returned by the Array.find function does not trigger a runtime error when trying to access the prefix property. This is called [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), and I'm pointing it out because I haven't been using this a lot a few months ago, but recently came across it again, and it's very handy (:

modules/nickname/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,67 @@
...
+
+    let user = await moduleModel.findOne({
+        attributes: ['userID', 'nickname'],
+        raw: true,
+        where: {
+            userID: newGuildMember.id
+        }
+    });
+    let memberName;
+    if (newGuildMember.nickname === null) {
```
SCDerox, line 32:
```suggestion
    if (!newGuildMember.nickname) {
```
In this code base, we usually do not compare to null and rather use this "shortcut". let me know if this is not suitable for this use case.

modules/nickname/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,67 @@
...
+
+    for (const role of roles) {
+        if (role.roleID === hoistrole) {
+            rolePrefix = role.prefix;
+        }
+    }
+
+    let user = await moduleModel.findOne({
+        attributes: ['userID', 'nickname'],
+        raw: true,
```
SCDerox, line 26: why are we doing this here? Usually leaving raw to false allows us to use sequelize's module interface and other features, while still allowing direct attribute access. (i don't really enjoy working with sequelize either anymore, prisma is much more fun)

modules/nickname/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,67 @@
...
+            userID: newGuildMember.id,
+            nickname: memberName
+        });
+
+    }
+
+    try {
+        await newGuildMember.setNickname(rolePrefix + memberName);
+    } catch (e) {
+    }
```
SCDerox, line 66: please add a log to ensure that our customer support can help users when they try doing stupid setups again

modules/nickname/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,67 @@
...
+    } else {
+        await moduleModel.create({
+            userID: newGuildMember.id,
+            nickname: memberName
+        });
+
+    }
+
+    try {
+        await newGuildMember.setNickname(rolePrefix + memberName);
```
SCDerox, line 64: We should probably make sure that the string isn't too long and maybe [truncate](https://github.com/ScootKit/CustomDCBot/blob/0689b9607c753148a619b391d842af5c670e8f38/src/functions/helpers.js#L263) it if it is

SCDerox (changes requested):
Thanks so much for opening this PR!

While reviewing your code (and without conducting testing), I have some questions and potential improvements. Please review them carefully, and please don't hesitate to voice your opinion if you disagree with me.
Thanks!

modules/nickname/events/guildMemberUpdate.js
hfgd123: Because I just copied it from somewhere else I used the database and it just works and I haven't tried it in this situation without raw. Will change if it works 👍

modules/nickname/events/guildMemberUpdate.js
hfgd123: where can I put the log for it to work as intended? just into console / return it / ...

modules/nickname/events/guildMemberUpdate.js
hfgd123: If it isn't it puts "undefined" in the rolePrefix String instead of leaving it how it was before (blank from initiation, so the user doesn't get a prefix at all)

modules/nickname/events/guildMemberUpdate.js
hfgd123: Done

modules/nickname/events/guildMemberUpdate.js
hfgd123: DOne

modules/nickname/events/guildMemberUpdate.js
hfgd123: Done

modules/nickname/events/guildMemberUpdate.js
SCDerox: You can use the log4js logger instance of the client, `client.logger`. So this would be `client.logger.error`

modules/nickname/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,68 @@
...
+module.exports.run = async function (client, oldGuildMember, newGuildMember) {
+
+    if (!client.botReadyAt) return;
+    if (newGuildMember.guild.id !== client.guild.id) return;
+    if (newGuildMember.nickname === oldGuildMember.nickname && newGuildMember.roles.highest.position === oldGuildMember.guild.me.roles.highest.position) return;
+
+    const roles = client.configurations['nickname']['config'];
+    const moduleModel = client.models['nickname']['User'];
+
+    let rolePrefix = "";
```
SCDerox, line 11:
```suggestion
```
declaration not required if [other review](https://github.com/ScootKit/CustomDCBot/pull/142/files#r1643483368) commit is commited, otherwise please ignore.

modules/nickname/events/guildMemberUpdate.js
SCDerox: Thanks for getting back to me. This is the case of poor type usage, but can be easily addressed by falling back to an empty string when no result has been found. I've updated by previous suggestion.

SCDerox (changes requested):

modules/nickname/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,53 @@
+const {truncate} = require("../../../src/functions/helpers");
+module.exports.run = async function (client, oldGuildMember, newGuildMember) {
+
+    if (!client.botReadyAt) return;
+    if (newGuildMember.guild.id !== client.guild.id) return;
+    if (newGuildMember.nickname === oldGuildMember.nickname && newGuildMember.roles.highest.position === oldGuildMember.guild.me.roles.highest.position) return;
```
SCDerox, line 6: `newGuildMember.roles.highest.position === oldGuildMember.guild.me.roles.highest.position`
Could you explain why we're comparing the highest role instead of the hoist role here (I'm not sure myself maybe I'm just stupid, but everywhere else we look at the hoist role)

SCDerox (changes requested):
Thanks so much for taking the time to address my previous review. I've tested your code and identified some additional small issues.

Additionally, I'd like to raise concerns regarding the limitation to the hoisted role: Right now, we only support hoisted roles in this module. I would suggest allowing every role (even ones that aren't hoisted) and parsing the prefix based on the highest role with prefix of the users. I think this would address use cases where user want to group members in the sidebar differently then their prefix. Here's an example: In a "Discord Team" there is a hierarchy based on roles in the team. Members will be grouped in the sidebar under "Staff", but everyone should have their team role as a role nickname (e.g. "Admin"). This isn't possible with this module right now. 

Additionally, I would like to suggest renaming the module (or the display name (what we call "human-readable name" for whatever reason) to "Role-Nicknames" to make its purpose more obvious.

modules/nickname/events/guildMemberUpdate.js
hfgd123: No, I can't. My plan what role to use changed while doing that and I forgot to change. But by switching to highest role now anyways it works, ig

hfgd123:
This would be a version with the highest role. The prolem: only the highest version. Not the highest one with a prefix. For that I would need to iterate through every role either in the config or the user's roles. Is there a simpler way to do that?

SCDerox:
Looping sounds like the easiest way in this case, but you can break out of the loop as soon as you've found the first role if you sort your array by role position

SCDerox:
Hi, just checking in, is this ready for review or do you have more work planned? :sweat_smile:

hfgd123:
Oh. Ready for review. I couldn't request it because it said you didn't yet finish your last one

modules/nicknames/models/User.js
```diff
@@ -0,0 +1,22 @@
...
+            tableName: 'nicknames_User',
+            timestamps: true,
+            sequelize
+        });
+    }
+};
+
+module.exports.config = {
+    'name': 'User',
+    'module': 'nickname'
```
SCDerox, line 21:
```suggestion
    'module': 'nicknames'
```

modules/nicknames/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,59 @@
...
+
+    if (!client.botReadyAt) return;
+    if (newGuildMember.guild.id !== client.guild.id) return;
+    if (newGuildMember.nickname === oldGuildMember.nickname && newGuildMember.roles.cache.size === oldGuildMember.guild.me.roles.cache.size) return;
+
+    const roles = client.configurations['nicknames']['config'];
+    const moduleModel = client.models['nicknames']['User'];
+
+    let rolePrefix = '';
+    let userRoles = newGuildMember.roles.cache.map(r => r.id);
```
SCDerox, line 12:
```suggestion
    let userRoles = newGuildMember.roles.cache.sort((a, b) => a.position - b.position).map(r => r.id);
```
Sorting by position ensures that the highest role gets added

SCDerox (changes requested):
So I found two small things, apart from that, everything looks good. The only thing I might want to suggest is checking that users have the correct role tags at startup, using a botReady event. When this event is triggered, we could fetch all members and loop through them, using the logic already implemented in guildMemberUpdate (simply export it as a function). This is just an idea on how to improve user experience of your module, let me know your thoughts on this.

SCDerox:
Thanks for taking the time to address my review. 
Looks great so far, just to confirm, do you want me to merge the PR as it is (as you requested) or are you planning to support the startup sync feature.

hfgd123:
I am working on it. The main process is now a separate function (in a separate file) and works like before again, but in the botReady event it tells me "not a function" and does nothing (require... is there). Still trying to fix that

SCDerox:
Let me know on discord if you need assistance ^^

modules/nicknames/events/guildMemberUpdate.js
hfgd123: Now botReady works but it gives me this error here:

TypeError: Cannot read properties of undefined (reading 'cache')
    at renameMember (D:\dev\WebstormProjects\CustomDCBot\modules\nicknames\renameMember.js:8:39)
    at module.exports.run (D:\dev\WebstormProjects\CustomDCBot\modules\nicknames\events\botReady.js:6:15)
    at Client.<anonymous> (D:\dev\WebstormProjects\CustomDCBot\main.js:413:103)
    at Client.emit (node:events:518:28)
    at startUp (D:\dev\WebstormProjects\CustomDCBot\main.js:193:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

SCDerox: To investigate this, I'd need access to your WIP files, please either perform a commit and push or send me both files on discord. Thanks.

modules/nicknames/events/botReady.js
```diff
@@ -0,0 +1,9 @@
+const {renameMember} = require('../renameMember');
+
+module.exports.run = async function (client) {
+    console.log("EEEE")
+    for (const member of client.guild.members.cache) {
```
SCDerox, line 5:
```suggestion
    for (const member of client.guild.members.cache.values()) {
```
The cache is a discord.js Collection, requiring the use of .values() to fix your issue

SCDerox (changes requested):

hfgd123:
Seems to be working now

modules/nicknames/renameMember.js
```diff
@@ -0,0 +1,56 @@
+const {truncate} = require("../../src/functions/helpers");
+
+renameMember = async function (client, guildMember) {
+    const roles = client.configurations['nicknames']['config'];
+    const moduleModel = client.models['nicknames']['User'];
+
+    let rolePrefix = '';
+    let userRoles = guildMember.roles.cache.sort((a, b) => a.position - b.position).map(r => r.id);
```
SCDerox, line 8:
```suggestion
    let userRoles = guildMember.roles.cache.sort((a, b) => b.position - a.position).map(r => r.id);
```
Actually, sorting is wrong sorry about this

modules/nicknames/renameMember.js
```diff
@@ -0,0 +1,56 @@
...
+    } else {
+        await moduleModel.create({
+            userID: guildMember.id,
+            nickname: memberName
+        });
+
+    }
+
+    try {
+        await guildMember.setNickname(truncate(rolePrefix + memberName, 32));
```
SCDerox, line 51: Two small things (sorry about not flagging this earlier):
* Could you add a small check to make sure that we're not setting nicknames that the user already has (check if the current nickname is the new nickname already)
* Could we add a check to not edit the nickname of the serverowner and add a (localized) client.logger.error call to ensure that customers are aware that the nickname of owners can't b edited?

modules/nicknames/renameMember.js
```diff
@@ -0,0 +1,56 @@
...
+    } catch (e) {
+        client.logger.error(e);
```
SCDerox, line 53: Please add some additional information (like "[nicknames] Could not edit nickname of %u. Error: %e", ideally localized)

SCDerox (changes requested):
Hi, thanks so much for implementing this feature.

Your PR is almost ready to be merged: If identified three potential issues regarding usability. Additionally, please ensure that you are complying with project standards by using ' instead of ".

modules/nicknames/events/guildMemberUpdate.js
```diff
@@ -0,0 +1,11 @@
+const {renameMember} = require('../renameMember');
+
+module.exports.run = async function (client, oldGuildMember, newGuildMember) {
+
+    if (!client.botReadyAt) return;
+    if (newGuildMember.guild.id !== client.guild.id) return;
+    if (newGuildMember.nickname === oldGuildMember.nickname && newGuildMember.roles.cache.size === oldGuildMember.guild.me.roles.cache.size) return;
```
SCDerox, line 7:
```suggestion
    if (newGuildMember.nickname === oldGuildMember.nickname && newGuildMember.roles.cache.size === oldGuildMember.roles.cache.size) return;
```
why would we compare the number of roles of the new guild member to the number of roles the client has on the guild?

modules/nicknames/renameMember.js
```diff
@@ -0,0 +1,63 @@
...
+            nickname: memberName
+        });
+
+    }
+
+    if (guildMember.nickname === truncate(rolePrefix + memberName, 32)) return;
+    try {
+        await guildMember.setNickname(truncate(rolePrefix + memberName, 32));
+    } catch (e) {
+        client.logger.error('[nicknames] ' + localize('nicknames', 'nickname-error', {u: guildMember.username, e: e}))
```
SCDerox, line 60:
```suggestion
        client.logger.error('[nicknames] ' + localize('nicknames', 'nickname-error', {u: guildMember.user.username, e: e}))
```
`guildMember.username` will always be undefined, only the User class has this property

modules/nicknames/renameMember.js
```diff
@@ -0,0 +1,63 @@
...
+
+renameMember = async function (client, guildMember) {
+    const roles = client.configurations['nicknames']['config'];
+    const moduleModel = client.models['nicknames']['User'];
+
+    if (guildMember.guild.ownerId === guildMember.id) {
+        client.logger.error('[nicknames] ' + localize('nicknames', 'owner-cannot-be-renamed', {u: guildMember.user.username}))
+        return;
+    }
+
```
SCDerox, line 12: So this will always display the message, could we move this when we have confirmed that the nickname does not meet the expectations?


```suggestion
```

modules/nicknames/renameMember.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+    } else {
+        await moduleModel.create({
+            userID: guildMember.id,
+            nickname: memberName
+        });
+
+    }
+
+    if (guildMember.nickname === truncate(rolePrefix + memberName, 32)) return;
```
SCDerox, line 56:
```suggestion
    if (guildMember.nickname === truncate(rolePrefix + memberName, 32)) return;
    if (guildMember.guild.ownerId === guildMember.id) {
       client.logger.error('[nicknames] ' + localize('nicknames', 'owner-cannot-be-renamed', {u: guildMember.user.username}))
       return;
    }
```

SCDerox (changes requested):

modules/nicknames/events/guildMemberUpdate.js
hfgd123: Because our friend autocomplete put things in there I didn't see

modules/nicknames/renameMember.js
```diff
@@ -0,0 +1,66 @@
...
```
SCDerox, line 51:
```suggestion
    if (memberName === truncate(rolePrefix + memberName, 32)) return;
```
we'd need to fall back to memberName to make this effective for users that do not have a nickname

SCDerox (changes requested):

modules/nicknames/renameMember.js
SCDerox, line 51:
```suggestion
    if (guildMember.displayName === truncate(rolePrefix + memberName, 32)) return;
```

We need to use the `displayName` property here to make this statement work for users without a nickname set.

SCDerox (changes requested):
Final thing :eyes:

hfgd123:
The break was to long :hAA: I forget many things

SCDerox (approved):
Thank you for your contribution and your time and kind cooperation during our review process. 
I'll go ahead and merge your PR. I'm excited to announce that your module will be available on SCNX shortly 🚀

SCDerox:
Thanks for your contribution - you have received a payout as part of our Open-Source Developer Program. You can find details in your SCNX organization payout history. To pay out the received amount, please follow instructions in your dashboard. If you have any questions, please contact ScootKit staff on scnx.app/help.

PR #158, opened 2025-05-21
fix[economy]: Fixed issue where perms aren't checked correctly

jateute:
Added an `await` to fix the perms


SCDerox (changes requested):

SCDerox (approved):
Thanks for the fast fix!

PR #168, opened 2025-12-15
Adds ping-protection module.

Kevinking500:
A full module dedicated to protecting specific users/roles for future pings!
Features:
- Sends a notice message with a GIF showing how to disable mention pings, about not being allowed to ping that user/role.
- User is able to configure the warnings message with embed functionality, while also being able to configure the protected roles, users, ignored roles and channels and to allow/don't allow reply pings
- User is able to configure if the bot will log ping history (default, yes and by default stored for 3 months.), if it logs moderation history (default no, as moderation capabilities is by default off. Stored for up to 12 months.) and if it keeps logs of leaved users (by default yes, by default up 1 day after leave, can be up to 7 days.)
- User is able to configure if the bot will mute or even kick the user after they reached xx amount of pings within a timeframe. By default it is 10 pings to trigger action within the configured timeframe of pings history logging. It also has an "advanced" configuration for that to allow for a custom timeframe in weeks. The mute duration can also be adjusted to liking.

Maybe even more features that I forgot to mention here lol
This module has been extensively tested by me and my testers over a 3 day period going over every feature, including after a full code polish. Changes may look longer/shorter in duration between them because I only commit changes once tested it works.

About AI Usage:
- I have used the following AI models:
> Gemini 3 pro
> Copilot
- These have been used for the following purpose:
> Auto completion (makes my life easier) (done by Copilot inside VSCode)
> Some debugging
> Explaining some things somewhat better (descriptions)


Kevinking500:
By the way to maybe answer some questions you may have:
- Why did I do select instead of integer in storage.json for database storage duration?
> I did select with the specified minimum and maximum amount of storage time to prevent user-side configuration errors
- Why is this not an add-on to the moderation module or ghost-ping module?
> This module is supposed to have many configuration fields, options and it wouldn't make sense to put it into a module, additionally I plan to add many other features in the future which would be inlogical to be inside a module
- Why did I not do a command warning for the panel command?
> The bot already checks if the executor has administration permissions and the delete data button checks for it again, as the executor is supposed to make it not in a public channel, but sometimes they do and therefore the data deletion button checks it again

Just in case you were wondering any of the above, then you have the answers ^^
(If any of the above require a change let me know)

locales/de.json
```diff
@@ -982,5 +982,43 @@
     "nicknames": {
         "owner-cannot-be-renamed": "Der Serverbesitzer (%u) kann nicht umbenannt werden.",
         "nickname-error": "Fehler beim Ändern des Nicknamens von %u: %e"
-    }
+    },
+    "ping-protection": {
+    "log-manual-delete": "Alle Daten für <@%u> (%u) wurden erfolgreich gelöscht.",
```
SCDerox: incorrect formatting

locales/en.json
```diff
@@ -1017,5 +1017,43 @@
     "nicknames": {
         "owner-cannot-be-renamed": "The owner of the server (%u) cannot be renamed.",
         "nickname-error": "An error occurred while trying to change the nickname of %u: %e"
-    }
+    },
+    "ping-protection": {
```
SCDerox, line 932: incorrect formatting

modules/ping-protection/commands/ping-protection-cmd.js
```diff
@@ -0,0 +1,131 @@
...
+                    description: localize('ping-protection', 'panel-description', { u: user.toString(), i: user.id }),
+                    color: 'BLUE',
+                    thumbnailURL: user.displayAvatarURL({ dynamic: true }),
+                    fields: [{
+                        name: localize('ping-protection', 'field-quick-history', {w: timeframeWeeks}),
+                        value: localize('ping-protection', 'field-quick-desc', { p: pingCount, m: modData.total }),
+                        inline: false
+                    }]
+                }]
+            });
```
SCDerox: embedType should not be used this way, it's intended to be used for user configuration data. Please provide a direct messageOption object or make the field configurable.

modules/ping-protection/commands/ping-protection-cmd.js
```diff
@@ -0,0 +1,131 @@
...
+
+            replyOptions.components = [row];
+            replyOptions.ephemeral = false;
```
SCDerox: use optionsToKeep in embedType call

modules/ping-protection/commands/ping-protection-cmd.js
SCDerox: This occurs in multiple places, please remove all incorrect usage.

modules/ping-protection/commands/ping-protection-cmd.js
SCDerox: This occurs in multiple places, please remove all incorrect usage.

modules/ping-protection/events/botReady.js
```diff
@@ -0,0 +1,21 @@
...
+const { localize } = require('../../../src/functions/localize');
+
+module.exports.run = async function (client) {
+    try {
+        await client.models['ping-protection']['PingHistory'].sync();
+        await client.models['ping-protection']['ModerationLog'].sync();
+        await client.models['ping-protection']['LeaverData'].sync();
+        
+    } catch (e) {
+    }
```
SCDerox: This is done by core, no need to do it manually.

modules/ping-protection/configs/moderation.json
```diff
@@ -0,0 +1,70 @@
...
+  },
+  "configElements": true,
+  "content": [
+    {
+      "name": "enableModeration",
+      "humanName": { "en": "Enable Moderation Actions", "de": "Bestrafungsaktionen aktivieren" },
+      "description": { "en": "If enabled, members who ping protected users/roles repeatedly will be punished.", "de": "Wenn aktiviert, werden Mitglieder, die geschützte Benutzer/Rollen wiederholt pingen, bestraft." },
+      "type": "boolean",
+      "default": { "en": false }
+    },
```
SCDerox, line 29: This field won't work if ping history is disabled, which is not made clear. Either make it clear to the user, that ping history is required for this to work or automatically enable ping history when this is enabled (and make that clear to the user as well)

modules/ping-protection/configs/moderation.json
```diff
@@ -0,0 +1,70 @@
...
+    "de": "Bestrafungsaktionen" 
+  },
+  "description": {
+    "en": "Define triggers for punishments.",
+    "de": "Definiere Auslöser für Bestrafungen."
+  },
+  "configElements": true,
+  "content": [
+    {
+      "name": "enableModeration",
```
SCDerox: This field toggles the whole config element, so please set the `elementToggle` property, instead of depending on this field for every other field in this file.

modules/ping-protection/configs/moderation.json
```diff
@@ -0,0 +1,70 @@
...
+      "humanName": { "en": "Enable Moderation Actions", "de": "Bestrafungsaktionen aktivieren" },
+      "description": { "en": "If enabled, members who ping protected users/roles repeatedly will be punished.", "de": "Wenn aktiviert, werden Mitglieder, die geschützte Benutzer/Rollen wiederholt pingen, bestraft." },
+      "type": "boolean",
+      "default": { "en": false }
+    },
+    {
+      "name": "advancedConfiguration",
+      "humanName": { "en": "Use Advanced Configuration", "de": "Erweiterte Konfiguration verwenden" },
+      "description": { "en": "If enabled, the advanced configuration below will be used and the basic configuration will be ignored.", "de": "Wenn aktiviert, wird die untenstehende erweiterte Konfiguration verwendet und die Basis-Konfiguration ignoriert." },
```
SCDerox: Not aquedacte enough. Describe differences between these options and assess wheter different options are even required (e.g. if it's enough to just have one option).

modules/ping-protection/configs/storage.json
```diff
@@ -0,0 +1,67 @@
...
+      "default": { "en": true }
+    },
+    {
+      "name": "pingHistoryRetention",
+      "humanName": { "en": "Ping History Retention", "de": "Ping-Verlauf Speicherzeit" },
+      "description": {
+        "en": "Decides on how long to keep ping logs. Minimum is 4 weeks (1 month) with a maximum of 24 weeks (6 months). This is the length factor of the 'Basic' punishment timeframe.",
+        "de": "Entscheidet, wie lange Ping-Logs gespeichert werden. Minimum 1 Woche, Maximum 24 Wochen. Dies ist der Faktor für den 'Basis'-Bestrafungszeitraum."
+      },
+      "type": "select",
```
SCDerox: Use integer and set maxValue / minValue.

modules/ping-protection/configs/storage.json
```diff
@@ -0,0 +1,67 @@
...
+      "content": [4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24],
+      "default": { "en": 12 },
+      "dependsOn": "enablePingHistory"
+    },
+    {
+      "name": "enableModLogHistory",
+      "humanName": { "en": "Enable Moderation Log History", "de": "Mod-Log Verlauf aktivieren" },
+      "description": { "en": "If enabled, the bot will keep a history of moderation actions taken by the ping-protection module.", "de": "Wenn aktiviert, speichert der Bot einen Verlauf der Bestrafungsaktionen des Ping-Schutz Moduls." },
+      "type": "boolean",
+      "default": { "en": false }
```
SCDerox: Should be probably enabled by default to ensure transparency of actions

modules/ping-protection/configs/storage.json
```diff
@@ -0,0 +1,67 @@
...
+    },
+    {
+      "name": "modLogRetention",
+      "humanName": { "en": "Moderation Log Retention (Months)", "de": "Mod-Log Speicherzeit (Monate)" },
+      "description": {
+        "en": "How long to keep records of punishments (1-12 Months).",
+        "de": "Wie lange Bestrafungsprotokolle gespeichert werden (1-12 Monate)."
+      },
+      "type": "select",
```
SCDerox: wrong type use

modules/ping-protection/events/guildMemberAdd.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Checks when a member rejoins the server and updates their leaver status
+ */
+
+const { markUserAsRejoined } = require('../ping-protection');
+
+module.exports.run = async function (client, member) {
+    if (!client.botReadyAt) return;
+    if (member.guild.id !== client.config.guildID) return;
```
SCDerox: Use `client.guildID` here (and for all other files doing this comparission)

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,113 @@
...
+    
+    // Identifies target
+    const targetUser = message.mentions.users.find(u => config.protectedUsers.includes(u.id));
+    const targetRole = message.mentions.roles.find(r => config.protectedRoles.includes(r.id));
+    const target = targetUser || targetRole;
+    const targetName = target.tag || target.name || target.id;
+    
+    // Checks if ping history logging is enabled
+    if (!storageConfig || !storageConfig.enablePingHistory) {
+        client.logger.info(`[ping-protection] User ${message.author.tag} pinged ${targetName}. Pings history logging is disabled, moderation actions cannot be done.`);
```
SCDerox: This is expected behaviour, as such no logging should take place.

```suggestion
```

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,113 @@
...
+    if (!client.botReadyAt) return;
+    if (!message.guild) return;
+    if (message.author.bot) return;
+    if (message.guild.id !== client.config.guildID) return;
+
+    const config = client.configurations['ping-protection']['configuration'];
+    const storageConfig = client.configurations['ping-protection']['storage'];
+    const moderationRules = client.configurations['ping-protection']['moderation'];
+    
+    if (!config || !moderationRules || !Array.isArray(moderationRules) || moderationRules.length === 0) return;
```
SCDerox:
```suggestion
    if (moderationRules.length === 0) return;
```
config and sotrageConfig values are ensured by core

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,113 @@
...
+    if (!pingedProtectedRole && !pingedProtectedUser) return;
+    
+    // Identifies target
+    const targetUser = message.mentions.users.find(u => config.protectedUsers.includes(u.id));
+    const targetRole = message.mentions.roles.find(r => config.protectedRoles.includes(r.id));
+    const target = targetUser || targetRole;
+    const targetName = target.tag || target.name || target.id;
+    
+    // Checks if ping history logging is enabled
+    if (!storageConfig || !storageConfig.enablePingHistory) {
```
SCDerox:
```suggestion
    if (storageConfig.enablePingHistory) {
```
correct value ensured by core

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,113 @@
...
+    
+    if (rule1.advancedConfiguration) {
+        requiredCount = rule1.pingsCountAdvanced;
+        generatedReason = localize('ping-protection', 'reason-advanced', { c: pingCount, w: rule1.timeframeWeeks });
+    } else {
+        requiredCount = rule1.pingsCountBasic;
+        generatedReason = localize('ping-protection', 'reason-basic', { c: pingCount, w: timeframeWeeks });
+    }
+    
+    client.logger.info(`[ping-protection] User ${message.author.tag} pinged ${targetName}. Count: ${pingCount}/${requiredCount}`);
```
SCDerox:
```suggestion
```

modules/ping-protection/models/LeaverData.js
```diff
@@ -0,0 +1,26 @@
+const { DataTypes, Model } = require('sequelize');
+
+module.exports = class LeaverData extends Model {
```
SCDerox:
```suggestion
module.exports = class PingProtectionLeaverData extends Model {
```
ensure module name is contained in class name to avoid sequelize issues, also name is not OOP

modules/ping-protection/models/ModerationLog.js
```diff
@@ -0,0 +1,39 @@
+const { DataTypes, Model } = require('sequelize');
+
+module.exports = class ModerationLog extends Model {
```
SCDerox:
```suggestion
module.exports = class PingProtectionModerationLog extends Model {
```
ensure module name is contained in class name to avoid sequelize issues, also name is not OOP

modules/ping-protection/models/LeaverData.js
SCDerox: Please fix for all other models as well

modules/ping-protection/ping-protection.js
```diff
@@ -0,0 +1,313 @@
...
+ * Logic for the Ping Protection module
+ * @module ping-protection
+ * @author itskevinnn
+ */
+const { Op } = require('sequelize');
+const { MessageActionRow, MessageButton } = require('discord.js');
+const { embedType, formatDiscordUserName, formatDate } = require('../../src/functions/helpers');
+const { localize } = require('../../src/functions/localize');
+
+const DISABLED_MSG = "History logging has been disabled by a bot-configurator.\nAre you (one of) the bot-configurators? You can enable history logging in the \"storage\" tab in the 'ping-protection' module ^^";
```
SCDerox: This should be localized.

modules/ping-protection/commands/ping-protection-cmd.js
```diff
@@ -0,0 +1,131 @@
+const {
```
SCDerox: filename should be command name

modules/ping-protection/commands/ping-protection-cmd.js
```diff
@@ -0,0 +1,131 @@
...
+// Main commands handler
+module.exports.run = async function (interaction) {
+    if (!interaction.guild) return;
+    
+    const group = interaction.options.getSubcommandGroup(false);
+    const subCmd = interaction.options.getSubcommand(false);
+    const config = interaction.client.configurations['ping-protection']['configuration'];
+    const isAdmin = interaction.member.permissions.has('ADMINISTRATOR') || 
+                    (interaction.client.config.admins || []).includes(interaction.user.id);
+// Handles subcommands
```
SCDerox: Subcommands should not be handled this way; use `module.exorts.subcommands` API (and `beforeSubcommands` if applibable)

SCDerox (changes requested):
Thanks for your contribution. Without testing your code, I was able to identify multiple issues that need to be addressed - feel free to comment on them if you disagree.

Additionally:

- Pings still go through, meaning that users still recive the pings "protected" by this module. This means that this module does not prevent pings, instead it punishes them. Instead of doing this, the module should create and maintain an [Auto Mod Rule](https://old.discordjs.dev/#/docs/discord.js/v13/class/AutoModerationRuleManager) and react to the [autoModerationActionExecution](https://old.discordjs.dev/#/docs/discord.js/v13/class/Client?scrollTo=e-autoModerationActionExecution) event when the rule gets triggered to run punishments. 

If you have any questions, concerns or disagree with any of my comments, please reply to them directly or let me know!

Happy holidays!

locales/de.json
Kevinking500: What exactly is incorrect in the formatting here?

locales/en.json
Kevinking500: What exactly is incorrect in the formatting here?

Kevinking500:
Thanks for the requested changes! I will immediately work on resolving these, for ones that I have questions about I let them as a reply.

modules/ping-protection/configs/storage.json
Kevinking500: I disabled it by default as moderation actions are also disabled by default, meaning if that's enabled it's useless unless moderation is turned on. That's a good notice though, i'll make it automatically turn on if moderation is enabled.

SCDerox: Okay, so if these features won't work without each other, we should probably use one toggle for both

SCDerox: Okay, so if these features won't work without each other, we should probably use one toggle for both

Kevinking500: it will still technically work, but won't log anything. When moderation actions is disabled, it will simply say no history found, while when disabled it will say it's disabled. When moderation actions is enabled, it will still moderate with this turned off, it's mainly to give users the option to decide to not log it and then they have full control over the logs

SCDerox: So in this case, this seems to be very confusing, so only one toggle for both options should be used since otherwise we just confuse users.

locales/de.json
SCDerox: JSON is incorrectly intended here; I suggest using the reformatting option in your IDE (I use JetBrains WebStorm and usually let it do the formatting before pushing). Additionally, the German translation contains errors, so you might want to remove it for now and we'll do it via Weblate 😄

Kevinking500: Alright. I am currently also sort of recoding the whole module because the new ways of works basically broke the module, so I have to get the new way of works with the new requirements. Do I remove German from the config files aswell or just from the locales? (Sorry if this is taking longer than you might expect)

Kevinking500:
Hey Simon, I have completed the changes requested. I think most (if not all) of your changes have been done, but it's been many hours so I can't remember if I did exactly all. Please let me know if everything has been solved now!

SCDerox (changes requested):
- Still not using automod APIs
- Please mark comments that you solved as "Resolved", otherwise I am unable to review the changes.

Kevinking500:
Alright, I just reviewed all changes and I did everything except indeed the automod API which I forgot, I will be working on that now!
Though the purpose was mainly to protect the user from future pings by informing the user they are not allowed to ping. This is exactly the behaviour as other bots, and if the ping gets deleted we have to remember that there might be important context provided which we then need to enter into that message. I think it's best to be an option as I think that people also want it to let people ping in for example important situations

SCDerox:
You could also add the automod feature as a configurable option, but I think to truly "protect" against pings, we need to actually be able to block them. This would make the module way more powerful :smile:

Kevinking500:
I definitely agree on that :)
I will add this as an option and enable it by default and it won't rely on moderation actions.
Thanks for that suggestion btw ^^

Kevinking500:
Hey, I have added that the bot will by default delete the original message with the ping, and adds the original content to the warning embed. Let me know if I need to adjust anything!
I didn't integrate Discord's native automod because it would overcomplicate it for a simple message deletion, though I might add it in the future if i want extra features from it.

SCDerox:
I think the Discord Native Automod has a crucial feature that you cannot do, even with deleting the message: blocking the message from triggering a ping to the user, so I would find it crucial to be supported rather than simply deleting messages.

Kevinking500:
Oh I actually didn't know that, that's a pretty neat feature! 
I definitely think that's very useful. Though one thing is that this module is meant to be my starting point with scnx and allowing me to get the basics before getting into the api stuff like we discussed, and therefore I think that this would be a great starting point without the automod and adding it later on after getting the full basics.  Now I am definitely not saying to wait months, just like get this published, see how it goes then apply api features.  Though if you think this is essential then I can try adding it now, though it will definitely take some longer to figure that out aswell. Don't mean this harshly towards you at all though, just a little notice ^^

SCDerox:
Thanks for explaining; however, I think it would be crucial to offer such a functionality at release, to ensure that we're delivering on user expectations. The automod API is built into discord.js (even at v13) and can be used easily (e.g. you just create a rule, update it every time the bot configuration changes, and listen to the event when it gets triggered), but I get  if you are not interested in implementing such functionality at this point. If that's the case, please let me know, and I will look into alternatives.

Kevinking500:
Ahhh okay if that's the case then I can definitely look into it a bit more for now. Thanks for providing that information as it's really useful, I think that if I do think that it's simple enough for now, that I will let you know tomorrow or atleast within the next 48 hours for a new commit with the automod integration ^^

SCDerox:
No stress, happy to help if you have any questions or need guideance - here on GitHub or in the #oss-chat on Discord :smile:

Kevinking500:
So I looked over the automod docs completely with rules, and I couldn't find something that let's the user send the ping message but make the ping not go through as an actual ping. Which action type is it that's needed or do you mean actually blocking the message?
Also now that I realised it, deleting the message feels pretty useless when we contain the actual message content in the embed as it's basically the same and the ping just results in a ghost ping, so either simply blocking the message all together, or somehow getting that ping suppression is a better choice than deleting I think

SCDerox:
So the easiest way to do this (the way we do it on our Discord as well), is to automatically sync the configured users as a keywword filter (trigger type: KEYWORD) with the following keywords for each user: `<@ID>` and `<@!ID>`.
This might look like this (with `<@!ID>` mentions missing): 
image
 When a message gets flagged by automod, Discord can block it automatically, by having an Action with Type BLOCK_MESSAGE configured (you can also set a configurable custom_message string).

Kevinking500:
Ah yes thanks for the additional information!
I have taken a look actually on a very popular bot Sapphire who also does something similar like this and sends a message with the content without the actual ping so the message is still sent by the bot without the actual ping, which I think is smart and I will take inspiration from that aswell. As when the message is blocked, the warning message and moderation actions features will not work as the message is not sent.
I will edit it to this, but isn't it supposed to be <@&ID> for roles? Or is <@!ID> something else?

SCDerox:
You can still trigger the warning message and moderation action manually, even without the message being sent by listing to the event and triggering relevant calls. 
Older client versions might use `<@!ID>` to mention users with usernames.

Kevinking500:
Yeah I am trying a workaround to log it into the history and moderate, I am planning a redesign of the module and logs when it is enabled, so it won't do something like showing a link to jump to the message when there is no such link etc.
I will let you know when it's ready ^^

Kevinking500:
I am currently facing multiple issues with it, currently thr only thing I can manage to get it done correctly is blocking the message. Other things seem to not work and are likely only going to work correctly in Discord.js V14 but I might add the commit of the code that blocks the message but doesn't do other things it's supposed to so you can take a look at it

SCDerox:
I am unsure what isn't working, your comment doesn't include any details. Since the API and event are fully supported on V13, simply upgrading discord.js won't likely address your issues.

Kevinking500:
It's mainly an issue with the reposting the message content and the alert channel for AutoMod.
When you enter in the ChannelID in your configuration for the alert channel, Discord seems to simply refuse it, even when making the ID passing it over to Discord as (ID) instead of "(ID)".
Also the bot is also unable to fetch the channel ID from the alert causing it to not know where to send the reposted message. I am currently focussing on primarily just the blocking mechanic of AutoMod, then commit after testing.
Additionally the bot seems to not be able to log it in the history either because of these 2 issues.
Please let me know if you know fixes for this. (Sorry that I don't provide error logs, I am already currently testing the simple mechanics and almost ready to commit the simple blocking)

SCDerox:
Please push your latest work, so I can take a look. I can't help without seeing your code.

Kevinking500:
Here you go, there might be some verbose logging, you can ignore that as I basically just reverted to an old version and I usually log alot to diagnose issues as I noticed some issues fail silently.

modules/ping-protection/events/autoModerationActionExecution.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+const { 
+    addPing, 
+    getPingCountInWindow, 
+    executeAction 
+} = require('../ping-protection');
+const { localize } = require('../../../src/functions/localize');
+
+module.exports.run = async function (client, execution) {
+    if (execution.ruleTriggerType !== 1) return;
```
SCDerox, line 5: This will be a string, as per docs: https://old.discordjs.dev/#/docs/discord.js/v13/typedef/AutoModerationRuleTriggerType, https://old.discordjs.dev/#/docs/discord.js/v13/class/AutoModerationActionExecution?scrollTo=ruleTriggerType

Kevinking500: Fixed, changed to KEYWORD

Kevinking500:
Damn that actually resolved a handful of issues.
After that the log channel did successfully get automatically set.
Though now we remain with the same issue of not able to find the channel ID of where the blocked message happened from the alert.
Also I accidentally forgot to complete the commit name and pressed commit :/
This is now the error in my terminal for the getting the channel ID for the reposting:
[WARN] [ping-protection] Repost Failed: Could not extract Channel ID

Kevinking500:
After some consideration I am doing the following changes instead:
I will make the alert channel required so it always sends the alert
I will add a custom message to the message block to let the user know that this message content has been sent to a channel where the user could see it
I will remove the repost ability as essentially the protected user/people with the role is already able to see it anyway.

I will commit the following changes after I coded and tested it ^^

SCDerox:
To get access to the channel, use [event].channelId (which needs to be fechted), to get access to the content, use [event].content

Kevinking500:
Ahhh okay noted. Though I removed it not only because I didn't get it to work, also because the alert message already shows the content to the protected people/people with the protected role meaning doing both a repost and sending the alert is basically double double. I have sent the full code of the bot now that works as supposed to!

Kevinking500:
Hi, i've found a quite critical bug when testing extensively today for Quality Assurance.
The allow reply pings configuration option did not work properly and the bot basically ignored it.
I have fixed this, tested it and confirmed it now works the following ways when it is set to allowed:
- The bot does not send a warning when a user replies to a protected user with mention enabled, UNLESS:
> It pings a protected user/role in the actual message which is checked, then it does send a warning, logs it and punishes if necessary.

Please review the full code with soon if you haven't already ^^ 
(or if you are currently testing it, you can safely copy this code into messageCreate.js as it's the only file edited and only changes the logic for checking reply pings)

Kevinking500:
Hey, I was bored and made it that instead of sending a warning message when a protected user pings themselves, that it will send a little meme and a chance to get a secret one

Kevinking500:
Hey, i've done some QOL improvements, which include all the following changes:
- Users can now configure the now renamed to "useCustomTimeframe" instead of "advancedConfiguration" in days instead of weeks
- The list commands have been merged (from /ping-protection list users/roles to /ping-protection list protected)
- The protected and whitelisted embeds have gone an overhaul - They have a new name, and a fresh description explaining what they are and do. Additionally instead of roles and users being right under eachother, they are now in fields with an inline meaning it provides better oversight
- The warning message has been updated to have a hyperlink to the list protected command (and I somehow also missed the 'command' word after 'use .....' in the previous versions)
- The "(basic configuration)" at the end of the reason for action history has been removed, and the "(Advanced configuration)" has been renamed to "(Custom timeframe)"
- There is now a bit more logging to provide better information, as those error logs are always silent without this.
- Probably some more I forgot to mention

I have tested this thoroughly since yesterday to make sure every feature works and for me it works as supposed to with no bugs ^^
(Happy new year btw!)

Kevinking500:
Hi, I've made a few changes I just came up with today thst I have no idea how I missed those features earlier:
- The configuration now also allows to whitelist specific users
- The config now allows the users to choose of they want to instead delete all pings history of an user after the time frame instead of 1 by 1.
- The config now also allows to choose id protected users can ping themselves or not, if they are not allowed to then they follow the usual procedure and if they are they get the chance to get rickrolled

I think that's about it ^^
Hope to hear back soon about the code :)

Kevinking500:
Whoops I accidentally made the logic opposite ways 😬,  fixed it though github is annoying and made me make 2 commits for 2 files changes

modules/ping-protection/commands/ping-protection.js
```diff
@@ -0,0 +1,144 @@
...
+  description: localize('ping-protection', 'cmd-desc-module'), 
+  usage: '/ping-protection',
+  type: 'slash',
+  options: [
+    {
+      type: 'SUB_COMMAND_GROUP',
+      name: 'user',
+      description: localize('ping-protection', 'cmd-desc-group-user'),
+      options: [
+        { type: 'SUB_COMMAND', name: 'history', description: localize('ping-protection', 'cmd-desc-history'), options: [{ type: 'USER', name: 'user', description: localize('ping-protection', 'cmd-opt-user'), required: true }] },
```
SCDerox: lines like this should be formatted to be more readable, e.g.
This applies to many parts of this module, I suggest running auto-format on the code with your IDE (i, personally, prefer Webstorm).

```suggestion
     {
            type: 'SUB_COMMAND_GROUP',
            name: 'user',
            description: localize('ping-protection', 'cmd-desc-group-user'),
            options: [
                {
                    type: 'SUB_COMMAND',
                    name: 'history',
                    description: localize('ping-protection', 'cmd-desc-history'),
                    options: [{
                        type: 'USER',
                        name: 'user',
                        description: localize('ping-protection', 'cmd-opt-user'),
                        required: true
                    }]
                },
                {
                    type: 'SUB_COMMAND',
                    name: 'actions-history',
                    description: localize('ping-protection', 'cmd-desc-actions'),
                    options: [{
                        type: 'USER',
                        name: 'user',
                        description: localize('ping-protection', 'cmd-opt-user'),
                        required: true
                    }]
                },
                {
                    type: 'SUB_COMMAND',
                    name: 'panel',
                    description: localize('ping-protection', 'cmd-desc-panel'),
                    options: [{
                        type: 'USER',
                        name: 'user',
                        description: localize('ping-protection', 'cmd-opt-user'),
                        required: true
                    }]
                }
            ]
        },
        {
            type: 'SUB_COMMAND_GROUP',
            name: 'list',
            description: localize('ping-protection', 'cmd-desc-group-list'),
            options: [
                {
                    type: 'SUB_COMMAND',
                    name: 'protected',
                    description: localize('ping-protection', 'cmd-desc-list-protected')
                },
                {
                    type: 'SUB_COMMAND',
                    name: 'whitelisted',
                    description: localize('ping-protection', 'cmd-desc-list-wl')
                }
            ]
        }
    ]
```

modules/ping-protection/commands/ping-protection.js
```diff
@@ -0,0 +1,144 @@
...
+  fetchModHistory, 
+  getPingCountInWindow,
+  generateHistoryResponse,
+  generateActionsResponse
+} = require('../ping-protection');
+const { localize } = require('../../../src/functions/localize');
+const { MessageActionRow, MessageButton, MessageEmbed } = require('discord.js');
+
+// Command definition
+module.exports.config = {
```
SCDerox: in this project we usually put the config object at the bottom of the file in every command

modules/ping-protection/events/autoModerationActionExecution.js
```diff
@@ -0,0 +1,60 @@
...
+    if (!isProtected) return;
+
+    let pingCount = 0;
+    let timeframeDays = 84;
+    let rule1 = (moderationRules && Array.isArray(moderationRules) && moderationRules.length > 0) ? moderationRules[0] : null;
+
+    if (!!storageConfig && !!storageConfig.enablePingHistory) {
+        const mockAuthor = { id: execution.userId };
+        const mockMessage = { author: mockAuthor, url: 'Blocked by AutoMod' };
+        const mockTarget = { id: rawId };
```
SCDerox: This might introduce issues in the future, e.g. when changes happen to the addPing function that require other parameters. Instead I suggest rewriting addPing to accept clearly named parameters instead of mock objects

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,152 @@
...
+    if (pingedProtectedUser) {
+        target = protectedMentions.first();
+    } else if (pingedProtectedRole) {
+        target = message.mentions.roles.find(r => config.protectedRoles.includes(r.id));
+    }
+
+    if (!target) return; 
+
+    // Funny easter egg when they ping themselves
+    if (target.id === message.author.id && config.allowSelfPing) {
```
SCDerox: While this behavior is fun, not all users want to have this. Please make this opt-in, e.g. the users need to expressly enable this and other easteregg features.

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,152 @@
...
+
+        const roll = Math.random();
+        let content = '';
+
+        if (roll < secretChance) {
+            content = secretMeme;
+            lastMemeMap.set(message.author.id, -1);
+            selfPingCountMap.delete(message.author.id); // Reset on secret unlock
+        } else if (currentCount === 5) {
+            content = 'Why are you even pinging yourself 5 times in a row? Anyways continue some more to possibly get the secret meme\n-# (good luck grinding, only a 1% chance of getting it and during testing I had it once after 83 pings)';
```
SCDerox: text should be localized

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,152 @@
...
+
+    if (!target) return; 
+
+    // Funny easter egg when they ping themselves
+    if (target.id === message.author.id && config.allowSelfPing) {
+        const secretChance = 0.01; // Secret for a reason.. (1% chance)
+        const standardMemes = [
+            '[Why are you the way that you are?](<https://www.youtube.com/watch?v=NY9UZI1OUMI>) - You just pinged yourself..',
+            '🔑 [Congratulations, you played yourself.](<https://www.youtube.com/watch?v=Lr7CKWxqhtw>)',
+            '🕷️ [Is this you?](<https://i.kym-cdn.com/entries/icons/original/000/023/397/C-658VsXoAo3ovC.jpg>) - You just pinged yourself.'
```
SCDerox: text should be localized

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,152 @@
...
+        requiredCount = rule1.pingsCountAdvanced;
+        generatedReason = localize('ping-protection', 'reason-advanced', { c: pingCount, d: rule1.timeframeDays });
+    } else {
+        requiredCount = rule1.pingsCountBasic;
+        const retentionWeeks = (storageConfig && storageConfig.pingHistoryRetention) ? storageConfig.pingHistoryRetention : 12; 
+        generatedReason = localize('ping-protection', 'reason-basic', { c: pingCount, w: retentionWeeks });
+    }
+
+    if (pingCount >= requiredCount) {
+        const { Op } = require('sequelize');
```
SCDerox: this should be included in the whole file not just here

SCDerox (changes requested):
While reviewing this code, I have noticed some things, please address them 😄 

As you might know, we're working on upgrading to the latest discord.js version. If you want, we can push the updated core and some modules to another branch and you can adjust your module to work with the latest version. This would allow us to accelerate your module availability on SCNX, since there's no need for us to manually upgrade it to the latest discord.js version.

Additionally, please disclose the following:

-  Please disclose whether Generative Large Language Models ("Generative AI", e.g. Coplilot) were used and to what extent to write this code. In addition, please confirm that you have reviewed and tested any generated code thoroughly. For details, please review our [AI Policy](https://corp.scootkit.com/docs/corp/internal/ai-usage/#6-ai-in-software-development-internal--oss).

Kevinking500:
Hey Simon, Thanks for the review!
I will address the following changes and include where and what I used AI for. Though one question about that, that should be in every file used right?
A different branch with the latest Discord.js version you are working on for the CB would be great so I can implement the new naming scheme directly and not crash the bot :)

SCDerox:
Disclosing how AI used as a comment here is fine or you can add it to your original PR body, no need to add it into the files themselves. 
Working on adding the new branch now.

SCDerox:
The branch is now available: [beta-discordjs14](https://github.com/ScootKit/CustomDCBot/tree/beta-discordjs14). Please [switch the base branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request) of this PR to that branch.

Kevinking500:
Hey Simon, i've addressed your feedback which you can see in my ping-protection module branch. I have not yet put itto the main branch here as I am currently working on adjusting the code to DJS V14, but in the meantime if you want to you can look there but after I changed the code to DJS V14 it will show up here anyway ^^
I will also quickly update the PR description about AI usage.

Kevinking500:
Alrighty, after alot of testing and things breaking, the Ping Protection module is finally done in Discord.JS V14, with very slight logic changes because of it (such as using .toJSON for actionRows etc because Discord.JS V14 basically requires it now)

Kevinking500:
I've also upgraded readability for the eyes in some files that I forgot earlier, I hope that helps aswell

Kevinking500:
Hi,
I've actually forgot a crucial feature I somehow forgot. These are the changes in this commit:
- Added action logs feature. The bot now sends a customisable message in the channel where the ping happened and the pinger got punished. This can be disabled.
- Improved readability for code even more
- Removed the resolveColor helper since I only now realised that I can just get the numbers for the color directly (thanks Gemini)
- Improved a few localization names to clarify better whether it's in the logs or in messages

modules/ping-protection/configs/moderation.json
```diff
@@ -0,0 +1,179 @@
...
+  "humanName": {
+    "en": "Moderation Actions"
+  },
+  "description": {
+    "en": "Define triggers for punishments."
+  },
+  "configElements": true,
+  "content": [
+    {
+      "name": "enableModeration",
```
SCDerox: Image

Image

This is confusing in the dashboard. This means when adding a moderation action, it will additonally need to be enabled in each config element. I don't think this is intended (since the easiest way to check if any moderation actions are required is to check the number of elements).

modules/ping-protection/configs/moderation.json
```diff
@@ -0,0 +1,179 @@
+{
+  "filename": "moderation.json",
+  "humanName": {
+    "en": "Moderation Actions"
+  },
```
SCDerox, line 5: consider adding a configElementName to improve UX in the dashboar: 
```
  "configElementName": {
    "en": {
      "one": "Punishment",
      "more": "Punishments"
    }
  },
```
Image

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,182 @@
...
+    if (pingedProtectedUser) {
+        target = protectedMentions.first();
+    } else if (pingedProtectedRole) {
+        target = message.mentions.roles.find(r => config.protectedRoles.includes(r.id));
+    }
+
+    if (!target) return; 
+
+    // Funny easter egg when they ping themselves
+    if (target.id === message.author.id && config.selfPingConfiguration === "Allowed, and ignored") return;
```
SCDerox: this configuration field (selfPingConfiguration) does not exist, only allowSelfPing is pushed

modules/ping-protection/commands/ping-protection.js
```diff
@@ -0,0 +1,206 @@
...
+  return module.exports.subcommands[sub](interaction);
+};
+
+// Handles subcommands
+module.exports.subcommands = {
+  'user': {
+    'history': async function (interaction) {
+      const user = interaction.options.getUser('user');
+      const payload = await generateHistoryResponse(interaction.client, user.id, 1);
+      await interaction.reply(payload);
```
SCDerox: replies should be ephemeral as per bot design standards (e.g. only the executing users should see it unless it is actionable by other users, for example in social settings)

modules/ping-protection/configs/storage.json
```diff
@@ -0,0 +1,97 @@
...
+      "type": "integer",
+      "default": {
+        "en": 12
+      },
+      "minValue": "4",
+      "maxValue": "24",
+      "dependsOn": "enablePingHistory"
+    },
+    {
+      "name": "DeleteAllPingHistoryAfterTimeframe",
```
SCDerox:
```suggestion
      "name": "deleteAllPingHistoryAfterTimeframe",
```
use camel case

modules/ping-protection/configs/configuration.json
```diff
@@ -0,0 +1,197 @@
...
+        "en": "AutoMod Log Channel" 
+      },
+      "description": { 
+        "en": "Channel where AutoMod alerts are sent." 
+      },
+      "type": "array",
+      "content": "channelID",
+      "default": { 
+        "en": [] 
+      },
```
SCDerox, line 157:
```suggestion
        "channelTypes": [   
        "GUILD_TEXT"
      ],
      },
```
add a channel types filter to ensure that logging is not attempted into voice channels.

modules/ping-protection/configs/configuration.json
```diff
@@ -0,0 +1,197 @@
...
+    {
+      "name": "autoModLogChannel",
+      "humanName": { 
+        "en": "AutoMod Log Channel" 
+      },
+      "description": { 
+        "en": "Channel where AutoMod alerts are sent." 
+      },
+      "type": "array",
+      "content": "channelID",
```
SCDerox:
```suggestion
      "type": "channelID",
```
only one log channel is supported

modules/ping-protection/ping-protection.js
```diff
@@ -0,0 +1,547 @@
...
+}
+
+// Syncs the native AutoMod rule based on configuration
+async function syncNativeAutoMod(client) {
+    const config = client.configurations['ping-protection']['configuration'];
+    
+    try {
+        const guild = await client.guilds.fetch(client.guildID);
+        const rules = await guild.autoModerationRules.fetch();
+        const existingRule = rules.find(r => r.name === 'SCNX Ping Protection');
```
SCDerox: We sell our Professional Plan under the promise of "offbrand", so we should consider naming this something without SCNX, e.g. "Ping Protection System" or someting (:

modules/ping-protection/ping-protection.js
```diff
@@ -0,0 +1,547 @@
...
+
+        // Logic to disable/delete the rule
+        if (!config || !config.enableAutomod) {
+            if (existingRule) {
+                await existingRule.delete().catch(() => {});
+            }
+            return;
+        }
+
+        const protectedIds = [...(config.protectedRoles || []), ...(config.protectedUsers || [])];
```
SCDerox: So this protects the role against being pinged, which is different from what you promise in the configuration field. Here, we should instead retrieve all members from cache (avoid fetching due to API rate limits for opcode 8) and add them to this array to correctly sync all members with the role.

modules/ping-protection/ping-protection.js
```diff
@@ -0,0 +1,547 @@
...
+                metadata: { channel: alertChannelId }
+            });
+        }
+
+        const ruleData = {
+            name: 'SCNX Ping Protection',
+            eventType: 1, 
+            triggerType: 1, 
+            triggerMetadata: {
+                keywordFilter: protectedIds.map(id => `*${id}*`)
```
SCDerox: While this works, this will also block users just typing the ID instead of actually pinging, which might be confusing. Instead, we should actually only block the pings:

```suggestion
                keywordFilter: [...protectedIds.map(id => `<@${id}>`), ...protectedIds.map(id => `<@!${id}>`) ]
```
you can also consider using a for loop to build this or building a regex which might be shorter than this rule.

SCDerox: based on the client only up to 1000 entries can be synced in a rule. consider adding a check if this object is over the limit.

modules/ping-protection/commands/ping-protection.js
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  await interaction.reply({ embeds: [embed.toJSON()] });
+}
+
+module.exports.config = {
+  name: 'ping-protection',
+  description: localize('ping-protection', 'cmd-desc-module'), 
+  usage: '/ping-protection',
+  type: 'slash',
```
SCDerox, line 165: defaultPermission should be set to avoid having this command availabile for everyone by default

modules/ping-protection/commands/ping-protection.js
```diff
@@ -0,0 +1,206 @@
...
+      ? config.protectedUsers.map(id => `<@${id}>`).join('\n') 
+      : localize('ping-protection', 'list-none');
+    
+    const rolesList = config.protectedRoles.length > 0 
+      ? config.protectedRoles.map(id => `<@&${id}>`).join('\n') 
+      : localize('ping-protection', 'list-none');
+
+    embed.addFields([
+      { name: localize('ping-protection', 'field-prot-users'), value: usersList, inline: true },
+      { name: localize('ping-protection', 'field-prot-roles'), value: rolesList, inline: true }
```
SCDerox: use truncate from the helper file to ensure that these values to not lead to invalid message payload errors

modules/ping-protection/commands/ping-protection.js
```diff
@@ -0,0 +1,206 @@
...
+      : localize('ping-protection', 'list-none');
+    
+    const channelsList = config.ignoredChannels.length > 0 
+      ? config.ignoredChannels.map(id => `<#${id}>`).join('\n') 
+      : localize('ping-protection', 'list-none');
+
+    embed.addFields([
+      { name: localize('ping-protection', 'field-wl-roles'), value: rolesList, inline: true },
+      { name: localize('ping-protection', 'field-wl-channels'), value: channelsList, inline: true }
+    ]);
```
SCDerox, line 152: use truncate from the helper file to ensure that these values to not lead to invalid message payload errors

modules/ping-protection/configs/configuration.json
```diff
@@ -0,0 +1,197 @@
...
+          "description": { 
+            "en": "ID of the pinged user/role"
+          }
+        }
+      ],
+      "default": {
+        "en": {
+          "title": "You are not allowed to ping %target-name%!",
+          "description": "<@%user-id%>, You are not allowed to ping %target-mention% due to your role. You can view which roles/members you are not allowed to ping by using the ping-protection list protected` command.\n\nIf you were replying, make sure to turn off the mention in the reply.",
+          "image": "https://c.tenor.com/zRR5cYKSIV0AAAAd/tenor.gif",
```
SCDerox: the dashoard might refuse to render this default value due to the external domain. consider uploading the gif to the SCNX User CDN.

modules/ping-protection/configs/configuration.json
```diff
@@ -0,0 +1,197 @@
...
+          "name": "target-id",
+          "description": { 
+            "en": "ID of the pinged user/role"
+          }
+        }
+      ],
+      "default": {
+        "en": {
+          "title": "You are not allowed to ping %target-name%!",
+          "description": "<@%user-id%>, You are not allowed to ping %target-mention% due to your role. You can view which roles/members you are not allowed to ping by using the ping-protection list protected` command.\n\nIf you were replying, make sure to turn off the mention in the reply.",
```
SCDerox: %user-id% is an undocumented paramter

modules/ping-protection/ping-protection.js
```diff
@@ -0,0 +1,547 @@
...
+            }
+            return `${(page - 1) * limit + index + 1}. **Pinged ${targetString}** at ${timeString}\n[Jump to Message](${entry.messageUrl})`;
+        });
+        description += lines.join('\n\n');
+    }
+
+    const row = new ActionRowBuilder().addComponents(
+        new ButtonBuilder()
+            .setCustomId(`ping-protection_hist-page_${userId}_${page - 1}`)
+            .setLabel('Back')
```
SCDerox: not localized, also below

modules/ping-protection/ping-protection.js
```diff
@@ -0,0 +1,547 @@
...
+            const reasonText = entry.reason || localize('ping-protection', 'no-reason') || 'No reason';
+            return `${(page - 1) * limit + index + 1}. **${entry.type}${duration}** - ${formatDate(entry.createdAt)}\n${localize('ping-protection', 'label-reason')}: ${reasonText}`;
+        });
+        description += lines.join('\n\n') + `\n\n*${localize('ping-protection', 'actions-retention-note')}*`;
+    }
+
+    const row = new ActionRowBuilder().addComponents(
+        new ButtonBuilder()
+            .setCustomId(`ping-protection_mod-page_${userId}_${page - 1}`)
+            .setLabel('Back')
```
SCDerox: not localized, also below

modules/ping-protection/ping-protection.js
```diff
@@ -0,0 +1,547 @@
...
+        description += localize('ping-protection', 'no-data-found');
+    } else {
+        const lines = history.map((entry, index) => {
+            const timeString = formatDate(entry.createdAt);
+            
+            let targetString = "Detected";
+            if (entry.targetId) {
+                targetString = entry.isRole ? `<@&${entry.targetId}>` : `<@${entry.targetId}>`;
+            }
+            return `${(page - 1) * limit + index + 1}. **Pinged ${targetString}** at ${timeString}\n[Jump to Message](${entry.messageUrl})`;
```
SCDerox: not localized, should check if messageurl is set or not (e.g. if it is a mock url)

modules/ping-protection/configs/configuration.json
```diff
@@ -0,0 +1,197 @@
...
+          "image": "https://c.tenor.com/zRR5cYKSIV0AAAAd/tenor.gif",
+          "color": "16711680"
```
SCDerox: the dashboard might not render this color due to its format.

```suggestion
          "color": "#ed4245"
```

modules/ping-protection/configs/configuration.json
```diff
@@ -0,0 +1,197 @@
...
+      "name": "autoModBlockMessage",
+      "humanName": { 
+        "en": "AutoMod custom message for message block" },
+      "description": { 
+        "en": "Custom text shown to the user when blocked (Max 150 characters)." 
+      },
+      "type": "string",
+      "maxLength": 150,
+      "default": { 
+        "en": "Protected User Ping: Your message was blocked but the content was sent to the log channel."
```
SCDerox: by default (unless configured), no message is being sent into a log channel, so we might want to consider rewording this

modules/ping-protection/configs/configuration.json
```diff
@@ -0,0 +1,197 @@
...
+    },
+    {
+      "name": "autoModBlockMessage",
+      "humanName": { 
+        "en": "AutoMod custom message for message block" },
+      "description": { 
+        "en": "Custom text shown to the user when blocked (Max 150 characters)." 
+      },
+      "type": "string",
+      "maxLength": 150,
```
SCDerox, line 171: this isn't supported by the dashboard, circling back on this

modules/ping-protection/models/PingHistory.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = class PingProtectionPingHistory extends Model {
+    static init(sequelize) {
+        return super.init({
+            userId: {
+                type: DataTypes.STRING,
+                allowNull: false
+            },
+            messageUrl: {
+                type: DataTypes.STRING,
+                allowNull: false
```
SCDerox: this should be true instead of using "mock urls". mock urls are very confusing to end users and shouldn't be used.

modules/ping-protection/commands/ping-protection.js
```diff
@@ -0,0 +1,206 @@
...
+      await interaction.reply(payload);
+    },
+    'panel': async function (interaction) {
+      const isAdmin = interaction.member.permissions.has('Administrator') || 
+                      (interaction.client.config.admins || []).includes(interaction.user.id);
+      
+      if (!isAdmin) return interaction.reply({ 
+        content: localize('ping-protection', 'no-permission'), 
+        flags: MessageFlags.Ephemeral 
+      });
```
SCDerox: This logic is confusing (since this would only allow trusted admins and administrators to use buttons while other users can use commands). The logic can be removed by making the command response ephemeral, outsourcing permission checks to Discord.

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,182 @@
...
+                content = localize('ping-protection', 'meme-grind');
+            } else {
+                const lastIndex = lastMemeMap.get(message.author.id);
+
+                let possibleMemes = standardMemes.map((_, index) => index);
+                if (lastIndex !== undefined && lastIndex !== -1 && standardMemes.length > 1) {
+                    possibleMemes = possibleMemes.filter(i => i !== lastIndex);
+                }
+
+                const randomIndex = possibleMemes[Math.floor(Math.random() * possibleMemes.length)];
```
SCDerox: use randomElementFromArray from the helpers file

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,182 @@
...
+module.exports.run = async function (client, message) {
+    if (!client.botReadyAt) return;
+    if (!message.guild) return;
+    if (message.guild.id !== client.guildID) return;
+
+    const config = client.configurations['ping-protection']['configuration'];
+    const storageConfig = client.configurations['ping-protection']['storage'];
+    const moderationRules = client.configurations['ping-protection']['moderation'];
+    
+    if (!config) return;
```
SCDerox: this bad practice, config (and other files) are always set by design by the core. consider removing checks like thisö

SCDerox (changes requested):
Hi, here are some additional issues I found. Please apologize for the delay, I am currently very sick sitting in a hotel room in Pitsburgh. Please let me know if you have any questions or concerns.
Additional feature suggestions:
- When a message is blocked by automod, consider adding an option that sends the warning message via DM to the user.

modules/ping-protection/configs/moderation.json
Kevinking500: This is almost intended behavior, I think I will remove the elementToggle and keep it correctly configured as I think it offers flexibility as to which exact punishments you want to have enabled/disabled at some time without having to delete the config element and re-configure and remember it at anytime when deciding you do actually want that.
I hope that's okay - otherwise I can think of a different solution (or remove it altogether)

modules/ping-protection/events/messageCreate.js
Kevinking500: I have no memory of what happened in configuration.json but apparently I accidentally deleted the selfPingConfiguration during editing once which I completely didn't notice during testing.
I apologise and will fix this immediately

modules/ping-protection/configs/configuration.json
Kevinking500: Didn't you say I had to do this lol
Anyways I will remove that then ^^

modules/ping-protection/configs/configuration.json
Kevinking500: Yes I understand, though also standardly we don't allow no log channel (allowNull = false, as standard) thus the reason I mentioned that, though I will reword this ^^

modules/ping-protection/configs/configuration.json
SCDerox: Yeah so this is mostly a note for me to check before publishing, all good, sorry about the confusion

modules/ping-protection/configs/moderation.json
SCDerox: I have to say, this is something that none of the other modules do and I think it might be rather confusing, in particular when adding new elements. 
Additionally, I'm having a hard time really seeing a use case here, I don't think it's that common to change punishments.

Kevinking500: I get your point, I have deleted that option and slightly reordered the positions of it for now. I am currently testing all the changes so it might take a bit until I commit it ^^

SCDerox: I would simply check if there are any config elements, another toggle in a different file would be very confusing.

SCDerox: I would simply check if there are any config elements, another toggle in a different file would be very confusing.

Kevinking500:
Hi Simon,
I have adjusted the code to the feedback and I believe I did everything. I clickied resolve on things I remember I did so it should be all good now!
I did adjust the logic of some code to properly allow multiple moderation actions as that wasn't tested before and didn't work properly.
Thanks for the feedback, have a great day and get better soon ❤️

modules/ping-protection/configs/configuration.json
```diff
@@ -148,9 +168,8 @@
         "en": "Custom text shown to the user when blocked (Max 150 characters)." 
       },
       "type": "string",
```
SCDerox, line 170:
```suggestion
      "type": "string",
      "maxLength": 150,
```

modules/ping-protection/commands/ping-protection.js
```diff
@@ -0,0 +1,214 @@
...
+        .setDescription(localize('ping-protection', 'panel-description', { u: user.toString(), i: user.id }))
+        .setColor('Blue')
+        .setThumbnail(user.displayAvatarURL({ dynamic: true }))
+        .addFields([{
+          name: localize('ping-protection', 'field-quick-history', { w: retentionWeeks }),
+          value: localize('ping-protection', 'field-quick-desc', { p: pingCount, m: modData.total }),
+          inline: false
+        }]);
+
+      await interaction.reply({ embeds: [embed.toJSON()], components: [row.toJSON()] });
```
SCDerox: still not ephemeral

modules/ping-protection/commands/ping-protection.js
SCDerox: This has not been resolved, even if it has been marked that way. In the latest push, there are still some responses that aren't ephemeral, in particular other subcommands.

modules/ping-protection/events/messageCreate.js
```diff
@@ -0,0 +1,181 @@
...
+    const moderationRules = client.configurations['ping-protection']['moderation'];
+
+    if (message.author.bot) return;
+
+    if (config.ignoredChannels.includes(message.channel.id)) return;
+    if (config.ignoredUsers.includes(message.author.id)) return;
+    if (message.member.roles.cache.some(role => config.ignoredRoles.includes(role.id))) return;
+
+    // Check for protected pings
+    const pingedProtectedRole = message.mentions.roles.some(role => config.protectedRoles.includes(role.id));
```
SCDerox, line 27: in the configuration description you promise users that "everyone with this role" will be protected. This only ensures that the role itself won't be pinged.

modules/ping-protection/events/autoModerationActionExecution.js
```diff
@@ -0,0 +1,84 @@
...
+    const isProtected = config.protectedRoles.includes(rawId) || config.protectedUsers.includes(rawId);
+    if (!isProtected) return;
+
+    if (!!storageConfig && !!storageConfig.enablePingHistory) {
+        try {
+            await addPing(client, execution.userId, null, rawId, false);
+        } catch (e) {}
+    }
+
+    if (!moderationRules || !Array.isArray(moderationRules) || moderationRules.length === 0) return;
```
SCDerox:
```suggestion
    if (moderationRules.length === 0) return;
```
both other checks are ensured by core and make the code less readable. Please remove all checks against configuration objects that are completed by core.

modules/ping-protection/ping-protection.js
```diff
@@ -0,0 +1,573 @@
...
+        const alertChannelId = getSafeChannelId(config.autoModLogChannel);
+        if (alertChannelId) {
+            actions.push({
+                type: 2, 
+                metadata: { channel: alertChannelId }
+            });
+        }
+
+        const ruleData = {
+            name: 'SCNX Ping Protection',
```
SCDerox:
```suggestion
            name: 'Ping Protection System',
```

SCDerox:
```suggestion
            name: 'Ping Protection System',
```

SCDerox (changes requested):
I have found some bugs and issues, please address them.
Additionally, when using the automod features, pings were not shown in the history command for me, could you check if this works for you? 
Thanks"

modules/ping-protection/ping-protection.js
Kevinking500: This is already done in the commit I recently did

Kevinking500: I think this is a duplicate as I don't think you mesnt to do the same twice, will resolve like the other one

modules/ping-protection/events/messageCreate.js
Kevinking500: I will reword that description to make it less confusing and that it only protects the role being pinged ^^

Kevinking500:
Thanks for the feedback!
I have been trying to diagnose the automod event not logging and I researched some more, also in the main.js code.
I have noticed the automoderationexecution intent was not in there, and after I added it for testing the logging worked perfectly.
If you think this would be fine to have in the main.js file, then it will be included in the next commit ^^
I am currently also diagnosing the following bugs I found during testing with AutoMod:
- Counting ping twice
- Logs show [Jump to message](Blocked by AutoMod) Which should simply be 'This ping was blocked by AutoMod'

modules/ping-protection/events/messageCreate.js
SCDerox: I think it would be better to actually protect the members of the role, similar to what you did with the auto mod 😁

Kevinking500: Sure ^^
I will add this soon

modules/ping-protection/commands/ping-protection.js
Kevinking500: All responses will now be ephemeral in the next commit

modules/ping-protection/events/messageCreate.js
Kevinking500: I have added this as an option and have it enabled by default which allows for even more customization ^^

Kevinking500:
Hey Simon, I have addressed the changes and it is now in this new commit.
- I have also polished and cleaned some code (which also makes ping-protection.js ~600 lines, but others way shorter)
- I have added a new intent to main.js as mentioned to make history logging with automod possible as that wasn't in the intents yet.

Thank you for the feedback - I haven't tested this version alot besides for the improvments/new features/slight changes - I will test this thoroughly soon

SCDerox:
Sounds great, please let me know when testing is completed.

Kevinking500:
Hi Simon, I have tested every feature to it's full extent after my initial testing and I noticed (and fixed) the following:
- Missing correct embed footers for the panel command and punishment error message
- Fixed the typo for the default warning message (was: using the ping-protection list protected` command, to: using the `/ping-protection list protected` comand)

Also added the most useless detail ever:
- Added an emoji to the why easter egg

So far I have found no bugs which is great ^^

SCDerox (approved):
Looking great to me, rolling out on SCNX Beta likely today!

PR #170, opened 2026-01-02
Some QOL improvements, including merging list commands and allowing for days instead of weeks in the now renamed "useCustomTimeframe" instead of "advancedConfiguration"

Kevinking500:


Kevinking500:
Whoops, accidentally sent it here instead of my own main branch

PR #171, opened 2026-01-30
Ping Protection V1.1

Kevinking500:
This is a small update to the current Ping Protection module.
This update includes the following changes:
- [QOL] Updated some locales and improved some of their names
- [Bug] Fixed the automod default block message from being over the 150 characters limit
- [Bug] Fixed that the whitelisted users didn't show

I apologize for not catching these bugs in the initial release
No AI has been used this time.


Fridolin012:
I noticed a few things I’d like to share regarding the module:
- Redundant Settings: In the "Moderation Actions" config, there’s a setting for "Pings to trigger moderation," but there’s also a custom function for it. It feels a bit messy to have both. I’d suggest just keeping the custom function ("custom timeframe") to make it cleaner.
- Log Retention: Why is the mod-log capped at 12 months? The moderation module doesn't have this limit, and I’m wondering if this can be changed to stay consistent with the rest of the tools.

Kevinking500:
Hi Fridolin, 
That is indeed a bit messy, I will change that.
For the log retention though, as this is a module for "only" pings and not otherwise a broad (multiple) rule breaking that can include a big variety of punishments like warnings, mutes, kicks, bans and things that it feels like having it more than 1 year for "only" pinging is a lot, but I can extend it if you feel like it would need to be more, or even do no limit.

Fridolin012:
Hi,

thank you for your quick response. I think we should ask @SCDerox for the second question^^

SCDerox:
Thanks for your valuable feedback, Fridolin. I, personally, have no issues with the limit. But based on your feedback, there's no technical reason that requires a limit. So making it adjustable and allowing users to disable it might be a compromise that both sides could agree on.

Kevinking500:
I think I might just increase the limit to 2 years (48 months) and if we get any feedback from users for it to be longer then ig I can make it unlimited. But I agree with both 😅

Kevinking500:
Alrighty, I have now set the pings history limit to 2 years (I actually relaized it was only 6 months max, that was indeed short). Additionally, I removed the pings count thing for the custom timeframe as that's basically useless, it now uses the standard one, which is now also renamed to pingsCount without the Basic.

Kevinking500:
Sorry, I forgot you were talking about moderation log

Kevinking500:
Hi, I quickly added the category feature in the configuration.json and storage.json config files to make it look more sleek ^^

Kevinking500:
Thanks to Simon who provided me the list of emoji's that actually work in the dashboard, I adjusted it to use supported ones ^^

SCDerox:
Hi, thanks so much for your contribution - quick update: Unfortunately I have exams this week, so it might take a few weeks to review everything. Thanks so much for your patience.

SCDerox (changes requested):
No issues from my side, please address the small thing copilot found and then re-request a review to get this merged :smile:

Kevinking500:
I have done the suggested changes, I will quickly be testing the moderation rules to make sure it works with these changes, and then i'll commit it ^^

Kevinking500:
I have found no bugs during the (fairly) short testing period of moderation actions ^^

Kevinking500:
I also addressed the tab in the locales I forgot to do which was mentioned by Jan in his latest PR here ^^

SCDerox (approved):

PR #183, opened 2026-03-11
Added name-list-cleaner AGAIN

hfgd123:
I don't like git


SCDerox (approved):
Looking great so far, going ahead with the merging.

PR #184, opened 2026-03-12
Adding option to specify whether they Move Channel and want to sync p…

ArgontTj:
…ermissions or not.

1. The Execution Logic (subcommands) We use the lockPermissions property in setParent to handle the sync.

2. The Command Configuration (config) Add the new BOOLEAN and STRING types to the options arrays.

Key Improvements :
- Permission Logic: Used { lockPermissions: sync }. When true, it behaves exactly like dragging a channel in the Discord UI and clicking "Sync Permissions."

- Audit Transparency: Every action now supports a reason which will appear in the Server Settings > Audit Log.

- Type Safety: I used integer types (e.g., type: 5 for Boolean) which are standard for raw object configurations in Discord.js, but ensured the logic remains clean.

### DIFF
--- a/modules/admin-tools/commands/admin.js
+++ b/modules/admin-tools/commands/admin.js
@@ -1,113 +1,54 @@
-const {ChannelType} = require('discord.js');
-const {localize} = require('../../../src/functions/localize');
+const { ChannelType } = require('discord.js');
+const { localize } = require('../../../src/functions/localize');
 
 module.exports.subcommands = {
     'movechannel': async function (interaction) {
         const channel = interaction.options.getChannel('channel', true);
-        if (!interaction.options.get('new-position')) return interaction.reply({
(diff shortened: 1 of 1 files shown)


SCDerox (changes requested):
Thanks for taking the time to submit this PR.

This code does not work since all command sync metadata has been removed. When submitting code, please ensure it works by testing it in a local environment.

Additionally, users overwriting the reason creates a dangerous lack of accountability. Instead of just using the  reason passed by the user, append the "Moved by" part to the reason to ensure admins can see in the log who moved a channel.

SCDerox:
It seems like this PR is stale. I'm going ahead and will be closing this for now. Please don't hesitate to comment or open a new PR when you've had time to address the concerns raised in the last review comment.

PR #193, opened 2026-05-28
Hotfix for Staff Management & Ping Protection modules

Kevinking500:
This PR consists of a hotfix for multiple bugs in my 2 modules - these bugs are fairly critical.
Besides it being a hotfix, it prepares the SMG module for the V2 development.

Bug fixes in this PR:
- [Ping protection]: Added a fallback behaviour if a localized modal key has more characters than Discord's limits, which causes the bot to not respond (currently)
- [Ping Protection] Added a '}' in the main logic file to fix a crash upon startup
- [Staff Management] Added a fallback behaviour if a localized modal key has more characters than Discord's limits, which causes the bot to not respond (currently)
- [Staff Management] Fixed that the reviews config didn't have an elementToggle, causing the module to be disabled internally if a reviews channel was not set (even if reviews are disabled)
- [Staff Management] Fixed incorrect staff permission checks that caused it to always decline permission unless the user has administrator perms (and raised it from staff > supervisor)
- [Staff Management] Added a staff permission check for promoting, as I somehow forgot that before allowing everyone to promote
- [Staff Management] Cleared JSON.parse/stringify because it's unnecessary
- [Staff Management] Changed tag from moderation > administration

I think this is a fairly urgent PR, but take your time as personal life is still the biggest priority ^^


Kevinking500:
Either Claude that reviewed previously was yapping about JSON.parse/stringify wasn't needed was lying, or copilot is lying. Though Copilot sounds more confident so I am trusting copilot here lol

PR #196, opened 2026-06-04
New module: Message-Quotes

JeanCoding16:
A quotes module that is extensively customizable. Aligned with the following suggestions: 
- https://featureboard.net/suggestions/6c040751-9534-4c98-80fd-a8aa4b6e5b47 
- https://featureboard.net/suggestions/47210fae-d5ca-4f50-be50-465de17173bf 

Featuring the following capabilities: 
- Role blacklist 
- Channel blacklist (Supports both specific channels and categories) 
- Include attachments? (Must be disabled for Components v2) 
- Do not quote bots? (Applies to links referencing bot messages; actual messages sent by bots are always ignored. Configurable; enabled by default for a more natural feel) 
- Reply to message? (Ignored if "Delete trigger message" is enabled) 
- Delete trigger message? 
- Message content is configurable (Using the following parameters: %userID%, %userName%, %userAvatar%, %channelID%, %channelName%, %timestamp%, %link%, %image%, %content%) 

Notes: The "Include attachments" option forwards images, videos, and GIFs; this may potentially require more granular filtering in the future.

### DIFF
--- a/modules/message-quotes/configs/config.json
+++ b/modules/message-quotes/configs/config.json
@@ -1,0 +1,108 @@
+{
+    "description": "Configure the message quoting system",
+    "humanName": "Configuration",
+    "filename": "config.json",
+    "content": [
+        {
+            "name": "roles",
+            "humanName": "Blacklist roles",
+            "description": "Roles that are excluded from quoting",
+            "default": [],
+            "type": "array",
+            "content": "roleID"
+        },
+        {
+            "name": "channels",
+            "humanName": "Blacklist channels",
+            "description": "Channels that are excluded from quoting (Channels and categories worked)",
+            "default": [],
+            "type": "array",
+            "content": "channelID"
+        },
+        {
+            "name": "withAttachments",
+            "humanName": "Attach files?",
+            "default": false,
+            "description": "Should all attachments be quoted? (Must be disabled for components v2!)",
+            "type": "boolean"
+        },
+        {
+            "name": "noBots",
+            "humanName": "Ignore bot messages?",
+            "default": true,
+            "description": "Bot messages are not included in the quote when activated",
+            "type": "boolean"
+        },
+        {
+            "name": "asReply",
+            "humanName": "Reply to messages?",
+            "default": true,
+            "description": "Reply to the message that triggered the quote (Ignored when \"Delete trigger\" is enabled)",
+            "type": "boolean"
+        },
+        {
+            "name": "deleteOrigin",
+            "humanName": "Delete trigger?",
+            "default": false,
+            "description": "When enabled, the trigger message will be deleted",
+            "type": "boolean"
+        },
+        {
+            "name": "message",
+            "humanName": "Message",
+            "description": "Message in which the quote is returned",
+            "default": {
+                "title": "Quote from #%channelName%",
+                "url": "%link%",
+                "description": ">>> %content%",
+                "image": "%image%",
+                "color": "#2ECC71",
+                "author": {
+                    "name": "%userName%",
+                    "img": "%userAvatar%"
+                }
+            },
+            "type": "string",
+            "allowEmbed": true,
+            "allowGeneratedImage": true,
+            "params": [
+                {
+                    "name": "%userID%",
+                    "description": "Od of the user"
+                },
+                {
+                    "name": "%userName%",
+                    "description": "Username of the user"
+                },
+                {
+                    "name": "%userAvatar%",
+                    "description": "Avatar of the user"
+                },
+                {
+                    "name": "%channelID%",
+                    "description": "Id of the channel from which the quote originates"
+                },
+                {
+                    "name": "%channelName%",
+                    "description": "Name of the channel from which the quote originates"
+                },
+                {
+                    "name": "%timestamp%",
+                    "description": "Shows when the original message was sent (Used discord timestamp)"
+                },
+                {
+                    "name": "%link%",
+                    "description": "Message-link of the original message"
+                },
+                {
+                    "name": "%image%",
+                    "description": "First image of the message, if available"
+                },
+                {
+                    "name": "%content%",
+                    "description": "Message content of the quote"
+                }
+            ]
+        }
+    ]
+}
--- a/modules/message-quotes/configs/config.json
+++ b/modules/message-quotes/configs/config.json
@@ -14,7 +14,7 @@
         {
             "name": "channels",
             "humanName": "Blacklist channels",
-            "description": "Channels that are excluded from quoting (Channels and categories worked)",
+            "description": "Channels that are excluded from quoting (Channels and categories are supported; a category excludes all its channels)",
             "default": [],
             "type": "array",
             "content": "channelID"
@@ -23,7 +23,7 @@
             "name": "withAttachments",
             "humanName": "Attach files?",
             "default": false,
-            "description": "Should all attachments be quoted? (Must be disabled for components v2!)",
+            "description": "Should all attachments be quoted? (Deactivation recommended if \"Message\" components v2 are used)",
             "type": "boolean"
         },
         {
@@ -31,6 +31,13 @@
             "humanName": "Ignore bot messages?",
             "default": true,
             "description": "Bot messages are not included in the quote when activated",
+            "type": "boolean"
+        },
+        {
+            "name": "selfQuote",
+            "humanName": "Allow Self-quotes?",
+            "default": true,
+            "description": "Can users quote their own messages?",
             "type": "boolean"
         },
         {
@@ -68,7 +75,7 @@
             "params": [
                 {
                     "name": "%userID%",
-                    "description": "Od of the user"
+                    "description": "Id of the user"
                 },
                 {
                     "name": "%userName%",
--- a/modules/message-quotes/configs/config.json
+++ b/modules/message-quotes/configs/config.json
@@ -82,6 +82,10 @@
                     "description": "Username of the user"
                 },
                 {
+                    "name": "%displayName%",
+                    "description": "Displays the user's nickname"
+                },
+                {
                     "name": "%userAvatar%",
                     "description": "Avatar of the user"
                 },
--- a/modules/message-quotes/events/messageCreate.js
+++ b/modules/message-quotes/events/messageCreate.js
@@ -1,16 +1,35 @@
-const { localize } = require('../../../src/functions/localize');
-const { embedType, embedTypeV2 } = require('../../../src/functions/helpers');
+const { 
+    embedType,
+    embedTypeV2,
+    formatDiscordUserName 
+} = require('../../../src/functions/helpers');
+const cooldowns = new Map();
 
 module.exports.run = async (client, msg) => {
     if (!client.botReadyAt) return;
+    if (!msg.content || msg.author.bot || msg.system) return;
     if (!msg.guild || !msg.member) return;
-    if (msg.author.bot || msg.system) return;
     if (msg.guild.id !== client.guildID) return;
     
-    const moduleConfig = client.configurations['message-quotes']['config'];
+    const now = Date.now();
+    const cooldownAmount = 5 * 1000;
+    if (cooldowns.has(msg.author.id)) {
+        const expirationTime = cooldowns.get(msg.author.id) + cooldownAmount;
+        if (now < expirationTime) return;
+    }
+    cooldowns.set(msg.author.id, now);
     
-    if (moduleConfig.channels.includes(msg.channel.id) || moduleConfig.channels.includes(msg.channel.parentId) || moduleConfig.channels.includes(msg.channel.parent?.parentId)) return;
-    if (msg.member.roles.cache.some(r => moduleConfig.roles.some(br => String(br) === r.id))) return;    
+    const moduleConfig = client.configurations['message-quotes']['config'] || {};
+    
+    const blacklistedChannels = moduleConfig.channels || [];
+    const blacklistedRoles = moduleConfig.roles || [];
+    
+    if (blacklistedChannels.includes(msg.channel.id) ||
+        blacklistedChannels.includes(msg.channel.parentId) ||
+        (msg.channel.parent?.parentId && blacklistedChannels.includes(msg.channel.parent.parentId))) {
+        return;
+    };
+    if (msg.member.roles.cache.some(r => blacklistedRoles.some(br => String(br) === r.id))) return;    
     
     const discordLinkRegex = /https:\/\/discord\.com\/channels\/(\d+)\/(\d+)\/(\d+)/i;
     const match = msg.content.match(discordLinkRegex);
@@ -23,6 +42,9 @@ module.exports.run = async (client, msg) => {
         const targetChannel = await msg.guild.channels.fetch(channelId).catch(() => null);
         if (!targetChannel || !targetChannel.isTextBased()) return;
         
+        const userPerms = targetChannel.permissionsFor(msg.member);
+        if (!userPerms || !userPerms.has('ViewChannel') || !userPerms.has('ReadMessageHistory')) return; 
+        
         const botPerms = targetChannel.permissionsFor(msg.guild.members.me);
         if (!botPerms || !botPerms.has('ViewChannel') || !botPerms.has('ReadMessageHistory')) return;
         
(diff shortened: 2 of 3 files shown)


JeanCoding16:
A description and an icon for the dashboard are missing.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+    
+    const [_, guildId, channelId, messageId] = match;
+    if (guildId !== msg.guild.id) return;
+    
+    try {
+        const targetChannel = await msg.guild.channels.fetch(channelId).catch(() => null);
+        if (!targetChannel || !targetChannel.isTextBased()) return;
+        
+        const botPerms = targetChannel.permissionsFor(msg.guild.members.me);
+        if (!botPerms || !botPerms.has('ViewChannel') || !botPerms.has('ReadMessageHistory')) return;
```
SCDerox, line 51: 🔴 **Security - blocker.** The bot's own `ViewChannel`/`ReadMessageHistory` is checked here, but not the triggering user's. Since message IDs are guessable snowflakes, any member can paste a link to a channel they can't read (staff, tickets, mod-log) and have its content + author + attachments reposted publicly. Could you add a check that the *member* can see the target channel too?

```js
const memberPerms = targetChannel.permissionsFor(msg.member);
if (!memberPerms || !memberPerms.has('ViewChannel') || !memberPerms.has('ReadMessageHistory')) return;
```

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+        const sendOptions = {
+            ...quoteMsg,
+            files: finalFiles.length > 0 ? finalFiles : undefined
+        };
+        
+        if (moduleConfig.asReply === true && moduleConfig.deleteOrigin !== true) {
+            sendOptions.allowedMentions = { repliedUser: false };
+            await msg.reply(sendOptions);
+        } else {
+            await msg.channel.send(sendOptions);
```
SCDerox, line 121: 🔴 **Security.** The `channel.send` path doesn't set `allowedMentions`, and the reply path (line 72) replaces it with just `{ repliedUser: false }`, dropping mention protection. Quoted content containing `@everyone`/`@role`/`@user` can fire live pings. Since we're echoing arbitrary user content, I'd suppress all mentions on both paths: `allowedMentions: { parse: [], repliedUser: false }`.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+        if (withAttachments && targetMsg.attachments.size > 0) {
+            targetMsg.attachments.forEach(att => {
+                files.push({
+                    attachment: att.url,
+                    name: att.name
+                });
+            });
+        }
+        
+        const firstAttachment = targetMsg.attachments.first()?.url || '';
```
SCDerox: 🔴 `attachments.first()?.url` is a raw Discord CDN URL with an expiry token - the image will 404 within a few hours, so quoted-image embeds break. Our `starboard` module handles this with `archiveDiscordAttachment()` (persistent storage) plus a regex fallback that scans message content for image URLs. Worth mirroring that here.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+
+module.exports.run = async (client, msg) => {
+    if (!client.botReadyAt) return;
+    if (!msg.guild || !msg.member) return;
+    if (msg.author.bot || msg.system) return;
+    if (msg.guild.id !== client.guildID) return;
+    
+    const moduleConfig = client.configurations['message-quotes']['config'];
+    
+    if (moduleConfig.channels.includes(msg.channel.id) || moduleConfig.channels.includes(msg.channel.parentId) || moduleConfig.channels.includes(msg.channel.parent?.parentId)) return;
```
SCDerox: 🔴 If `channels`/`roles` are ever undefined (e.g. first boot before defaults are written), `.includes()`/`.some()` throw here - and this runs *before* the try/catch. Mind wrapping them as `(moduleConfig.channels || [])` / `(moduleConfig.roles || [])`? (Also, `msg.channel.parent?.parentId` silently misses a blacklisted category if the parent isn't cached - minor, but noting it.)

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+    if (msg.member.roles.cache.some(r => moduleConfig.roles.some(br => String(br) === r.id))) return;    
+    
+    const discordLinkRegex = /https:\/\/discord\.com\/channels\/(\d+)\/(\d+)\/(\d+)/i;
+    const match = msg.content.match(discordLinkRegex);
```
SCDerox, line 35: 🟡 `msg.content` can be `null` on partial messages, and this runs before the try/catch, so it'd throw unhandled. A quick `if (!msg.content) return;` up top covers it. Also worth a per-user cooldown around here - right now there's nothing stopping a user from pasting many links rapidly and triggering a fetch+send storm.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+        
+        const targetMsg = await targetChannel.messages.fetch(messageId).catch(() => null);
+        if (!targetMsg) return;
+        
+        if (moduleConfig.noBots === true && targetMsg.author.bot) return;
+        
+        let files = [];
+        const withAttachments = moduleConfig.withAttachments;
+        if (withAttachments && targetMsg.attachments.size > 0) {
+            targetMsg.attachments.forEach(att => {
```
SCDerox: 🟡 No cap on attachment count/size - quoting a message with 10×25 MB files re-uploads all of them. A small limit (e.g. cap at 1-3, skip oversized) would keep this safe.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+        };
+        
+        if (moduleConfig.asReply === true && moduleConfig.deleteOrigin !== true) {
+            sendOptions.allowedMentions = { repliedUser: false };
+            await msg.reply(sendOptions);
+        } else {
+            await msg.channel.send(sendOptions);
+        }
+        
+        if (moduleConfig.deleteOrigin === true) {
```
SCDerox, line 124: 🟡 When `deleteOrigin` is on but the bot lacks `ManageMessages`, the quote posts but the trigger silently stays, with no log. A `client.logger.warn(...)` in the else branch would make that debuggable.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+                });
+            });
+        }
+        
+        const firstAttachment = targetMsg.attachments.first()?.url || '';
+        const userAvatar = targetMsg.author.displayAvatarURL({ dynamic: true });
+        const unixSeconds = Math.floor(targetMsg.createdTimestamp / 1000);
+        const quoteMsg = await embedTypeV2(moduleConfig.message, {
+           '%userID%': targetMsg.author.id,
+           '%userName%': targetMsg.author.tag,
```
SCDerox: 🟢 `author.tag` is the legacy `name#0000` format (returns just the username for migrated accounts). We have a `formatDiscordUserName()` helper in `src/functions/helpers.js` for this. Also, consider exposing the author's *server nickname* (`%displayName%`) and a `%channelMention%` param like `starboard` does - admins often want those in the embed.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+            targetMsg.attachments.forEach(att => {
+                files.push({
+                    attachment: att.url,
+                    name: att.name
+                });
+            });
+        }
+        
+        const firstAttachment = targetMsg.attachments.first()?.url || '';
+        const userAvatar = targetMsg.author.displayAvatarURL({ dynamic: true });
```
SCDerox: 🟢 `displayAvatarURL({ dynamic: true })` - `dynamic` was removed in discord.js v14 (animated avatars resolve by default now), so this option is just ignored. Can drop it.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
...
+        if (moduleConfig.noBots === true && targetMsg.author.bot) return;
+        
+        let files = [];
+        const withAttachments = moduleConfig.withAttachments;
+        if (withAttachments && targetMsg.attachments.size > 0) {
+            targetMsg.attachments.forEach(att => {
+                files.push({
+                    attachment: att.url,
+                    name: att.name
+                });
```
SCDerox, line 70: 🟢 `att.name` can be `null` (e.g. voice messages); `name: att.name ?? 'attachment'` avoids an edge-case upload error.

modules/message-quotes/events/messageCreate.js
```diff
@@ -0,0 +1,87 @@
+const { localize } = require('../../../src/functions/localize');
```
SCDerox: 🟢 `localize` is imported but never used - since the embed is fully admin-configured there's nothing to localize here, so this can be removed.

SCDerox (commented):
Hey @JeanCoding16 👋 thanks a lot for this - a quotes module has been one of the most-requested features on the board, and you clearly put real care into it: the config surface is thoughtful, the permission checks on the target channel are a nice touch, and the attachment/reply/delete options cover a lot of real use-cases. Really appreciate the contribution. 🙌

I went through it in depth and left inline comments. Heads up: **this was an AI-assisted review**, so a human has sanity-checked the important points, but please push back on anything that looks off - a few are judgment calls rather than hard rules.

There's **one security issue I'd consider a blocker** before merge, plus a handful of smaller must-fixes, some "would be great to have" config ideas, and a few typos. Grouped overview:

**🔴 Must-fix before merge**
- **Cross-channel info disclosure (security):** the bot checks whether *it* can read the linked channel, but not whether the *user who pasted the link* can. Right now any member can paste a link to a staff/ticket/mod channel and have its contents reposted into a public channel. This is the big one - details inline.
- **Mention re-pinging (security):** quoted content can fire live `@role`/`@user` pings (and `@everyone` on the reply path). Suppress mentions on both send paths.
- **Broken icon:** `fas fa-chat` isn't a valid icon and won't render.
- **Expiring image links:** quoted images use raw Discord CDN URLs that 404 after a few hours.
- **Crash if the config arrays aren't set yet** (first boot before defaults persist).

**🟡 Should-fix / robustness**
- No rate-limit or cooldown - a single user can spam quotes and burn the bot's API budget.
- Silent failure when "delete trigger" is on but the bot lacks Manage Messages (no log).
- Attachment forwarding has no count/size cap.
- Empty `>>> ` blockquote when quoting an image-only message.

**🟢 Nice-to-have config ideas** (totally your call - the module works without these)
- A channel *allow-list* (not just blacklist), a `selfQuote` toggle, a per-user cooldown, and an optional trigger keyword so quoting is opt-in rather than firing on every link. These mirror what our `starboard` module offers.
- Surfacing the original message's timestamp in the default embed, and exposing the author's server nickname + a channel mention as params.

**✍️ Typos / copy** (these show up in the dashboard, so worth a quick pass) - flagged inline.

Thanks again - this is genuinely close, and the security fix aside it's mostly polish. Happy to clarify any of the comments. 🚀

modules/message-quotes/configs/config.json
```diff
@@ -0,0 +1,108 @@
...
+            "type": "boolean"
+        },
+        {
+            "name": "message",
+            "humanName": "Message",
+            "description": "Message in which the quote is returned",
+            "default": {
+                "title": "Quote from #%channelName%",
+                "url": "%link%",
+                "description": ">>> %content%",
```
SCDerox, line 64: 🟡 The default `">>> %content%"` renders as a bare `>>> ` blockquote when the quoted message has no text (image-only/sticker). Worth falling back to something, or dropping the `>>> ` prefix. Also, `%timestamp%` is passed to the embed but isn't used in this default - adding it (e.g. in the footer) would preserve the original send time, which otherwise gets lost.

modules/message-quotes/configs/config.json
```diff
@@ -0,0 +1,108 @@
...
+            "humanName": "Blacklist roles",
+            "description": "Roles that are excluded from quoting",
+            "default": [],
+            "type": "array",
+            "content": "roleID"
+        },
+        {
+            "name": "channels",
+            "humanName": "Blacklist channels",
+            "description": "Channels that are excluded from quoting (Channels and categories worked)",
```
SCDerox: ✍️ Small copy fix: "Channels and categories worked" reads a bit broken - maybe "Channels and categories are supported (a category ID excludes all its channels)".

modules/message-quotes/configs/config.json
```diff
@@ -0,0 +1,108 @@
...
+            "default": [],
+            "type": "array",
+            "content": "channelID"
+        },
+        {
+            "name": "withAttachments",
+            "humanName": "Attach files?",
+            "default": false,
+            "description": "Should all attachments be quoted? (Must be disabled for components v2!)",
```
SCDerox: ✍️ "(Must be disabled for components v2!)" is a bit internal - could reword as admin-facing guidance about when to turn this off.

modules/message-quotes/configs/config.json
```diff
@@ -0,0 +1,108 @@
...
+                    "img": "%userAvatar%"
+                }
+            },
+            "type": "string",
+            "allowEmbed": true,
+            "allowGeneratedImage": true,
+            "params": [
+                {
+                    "name": "%userID%",
+                    "description": "Od of the user"
```
SCDerox: ✍️ Typo: "Od of the user" → "ID of the user".

modules/message-quotes/module.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "message-quotes",
+  "humanReadableName": "Message quotes",
+  "description": "",
+  "fa-icon": "fas fa-chat",
```
SCDerox: 🔴 `fas fa-chat` isn't a valid icon in our set, so it won't render. For a quotes module `fas fa-quote-left` fits nicely (`fas fa-comments` / `fas fa-message` also work). Full list of supported icons: https://scnx.app/developers/icons

modules/message-quotes/module.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "message-quotes",
+  "humanReadableName": "Message quotes",
+  "description": "",
```
SCDerox: ✍️ `description` is empty - a one-liner here shows up in the module browser, e.g. "Quotes a Discord message when a user pastes a message link."

SCDerox:
One more pointer @JeanCoding16 - while you're tweaking `config.json` (the copy fixes and any of the new config ideas), our config-file docs are handy for previewing how each option renders in the dashboard: https://scnx.app/developers/configuration 👍

JeanCoding16:
I've changed the points mentioned. I can't view this page; I'm missing the flag: https://scnx.app/developers/configuration. And regarding `archiveDiscordAttachment()`: that would need to be added separately. I've taken a few steps for it, but I can't implement it in a way that allows me to guarantee the functionality.

modules/message-quotes/events/messageCreate.js
JeanCoding16: Instead of offering `channel mention` as a separate parameter, I thought of using `<#%channelID%>`, but mentioning the channel if it doesn't lead to the message is rather misleading. That's why I later removed it as a parameter.

modules/message-quotes/configs/config.json
JeanCoding16: I have implemented fallback values such as: [Attachment], [Sticker] and [None]. This should ensure that the formatting still works and the user receives a info about the quoted message.

SCDerox:
Thanks for the quick turnaround @JeanCoding16. The two security fixes (the member permission check and `allowedMentions` on both send paths) look correct, and the attachment cap, cooldown, and content fallbacks all work. A few follow-ups below. As before, this was an AI-assisted review, so feel free to push back on anything.

**On the persistent images** - this is simpler than it looked, so I don't think you need to leave it out. `archiveDiscordAttachment(client, url, meta)` in `src/functions/helpers.js` always returns a usable URL: when archival is available it returns a permanent storage URL, and on any failure it returns the original Discord URL unchanged. That covers OSS builds where the storage integration isn't shipped, archival being disabled, rate limits, quota, and upstream errors. So it's a safe drop-in: no worse than what you have now, and persistent automatically when the bot runs with storage enabled. (There's also `tryArchiveDiscordAttachment` if you ever need to know whether it actually persisted, since it returns `null` instead of the original URL, but here you want the always-returns-a-URL one.)

`starboard` already does this if you want a reference: `modules/starboard/handleStarboard.js`, the `archiveDiscordAttachment(...)` call plus the `msg.content.match(...)` image fallback just below it. Applied here it would be roughly:

```js
let finalImage = '';
const firstAttachment = targetMsg.attachments.first();
if (firstAttachment) {
    finalImage = await archiveDiscordAttachment(client, firstAttachment.url, {
        displayName: `Quote by ${formatDiscordUserName(targetMsg.author)} in #${targetChannel.name}`.slice(0, 100),
        tags: ['message-quotes'],
        uploaderDiscordID: targetMsg.author.id
    });
} else {
    // covers messages that are just a pasted image link
    const imgMatch = targetMsg.content.match(/https?:\/\/\S+\.(?:png|jpe?g|gif|webp)/i);
    if (imgMatch) finalImage = imgMatch[0];
}
```

That also covers messages that are just a pasted image link with no attachment.

**On the cooldown**, one thing to check: `cooldowns.set(...)` currently runs before the link regex, so every message a user sends starts their 5s timer. If someone sends a normal message and then pastes a quote link within 5s, the quote won't fire. Moving the cooldown check and set to after `if (!match) return;`, so it only counts actual quote triggers, should fix it.

Smaller stuff:
- `module.json` description has a typo: "when a user has pastes a message link" should be "when a user pastes a message link".
- `%displayName%` is now passed to the embed but isn't listed in the `params` array in `config.json`, so admins won't see it in the dashboard. Worth adding an entry.
- Optional: `%timestamp%` still isn't used in the default embed, so the original send time isn't shown by default. Adding it to the footer would be a nice touch, but isn't required.

Also, you should have access to the config preview docs now (https://scnx.app/developers/configuration), so previewing the new param and copy changes should work. Sorry about the flag gap earlier.

The `%channelMention%` reasoning and the `[Attachment]/[Sticker]/[None]` fallbacks both make sense. Once the cooldown ordering and the image helper are in, I think this is about there.

JeanCoding16:
Fixed. I haven't found anything yet on how to replace the footer timestamp with %timestamp%, so it's still not included.

PR #204, opened 2026-07-25
Staff Management System V2

Kevinking500:
# Staff Management System V2 - A big overhaul of many features /w big improvements

This **planned** update will contain multiple completely reworked features. This message will be big, just a small warning. To say the least, it's a big overhaul of many features in this module, to make the experience way better.

**Changelogs**
- **Rework on the staff-status request;** the command now also lets users set a date (DD-MM-YYYY) when the status is supposed to start. This also comes with the improvement to ending the status after the duration from the request time, instead of approval time.
- **Status reminders:**  A configurable option in the dashboard allows the bot to send a reminder x hours before the staff member's status expires. This is sent in the DM's as a customizable message.
- **Status panel:** The configuration allows to send a customizable message in a configurable channel, which shows who are currently on LoA/RA. It's seperate for each status type. The embed also shows buttons to start a request and view their current staff-status status.
- **Full rework of the activity checks system:**
> AC reminders: The configuration now allows the option to send a reminder message x hours before the activity check ends to remind staff to respond
> The configuration now allows a 'view respondents' button to be shown on the AC message. This list shows the total amount of respondents + the mention of the users in the embed.
> The AC view command will now be more informational, with things like the full respondents list (button), total amount of respondents, total amount of staff who didn't respond, exempted people and more.
> The bot now replies with a link to the AC message instead of simply the channel where the AC message is sent. This applies to, for example, the activity-check start command.
> Activity Check ID's can be used as a parameter instead of only being used internalyl. This also adds the 'Activity Check ID: xx' field to the default activity check messages for recognition. This will also allow specific activity check lookups via the ID of that check
- **Shifts system overhaul:**
> The duty admin panel now allows the removal of duty time.
> A new `/duty server-panel` command or similar will allow management+ and administrators manage multiple things for the whole server. This includes wiping duty time for the whole server, voiding all shifts, forcing everyone off-duty, and transferring everyone onto 1 shift type. Wiping the duty time for everyone in the server comes with a 24hr cooldown after it happened.
> Going on break is completely reworked. The configuration now allows full customization on it: Max break duration, Max total break duration (over the whole shift), max break count, Minimum on-duty time before a break and break cooldown.
> EOS report is getting more detailed, with a 'show detailed report' button, showing more exact details. These details include things like the exact starting/end time, exact start/end time of each break they took etc.
> The configuration also allows to automatically end a shift after x hours - this is very useful to prevent shift farmers.
> The configuration now also allows a duty shifts panel - this sends a customizable panel into a configurable channel to show who are currently on-duty for each shift type. This embed also shows the normal duty buttons, such as going on duty, on break, or going off duty. This embed automatically updates every 5 minutes, or when a shift change happens. From the shift type, there is another 5 minute countdown.
- **Reviews overhaul:**
> Reviews can be deleted by supervisor/management+ for invalid reviews
> A notification can be set so the staff member can get a notification in DM's as well when they get reviewed
> Option to automatically create a thread on the review's message to discuss the review
> A new command to block a specific user from reviewing staff members
> An option to allow staff members to "say thanks" via a button on the review, which sends a custom "thanks!" message to the author's DM's
- **Automated termination:** Allows an option for an automated termination (infraction name = termination), which removes all roles, including, and above the staff base role or the configured staff role. Unlike suspensions, this is not tied to a duration. This can be heavily abused for raids, so it's off by default with a notice.
- **Optional safeguard for infractions:** The configuration now has an option (default = true) to have a safeguard, which prevents a user from infracting a person with a role higher than them. This is especially great for preventing raids, especially for suspensions and automatic terminations.
- **Improved data deletion safety:** Just like in Ping protection, the data deletion is receiving a small safety update. A cooldown is being introduced for deleting data for an individual feature - this cooldown is 24hr and only applies to that user. When all data for that user is deleted, a cooldown of 168hrs starts - this also only applies to that user and is not sever-wide for all users. Additionally, the staff member who deleted data, goes on a 24hr cooldown if they deleted all data for an user.
- **Anti-raid config:** This is an effort to prevent staff abuse internally directly from this module. It integrates with promotions, infractions and other (dangerous) actions to prevent a server raid before it's noticed by other bots. This will also (depending on the configuration) block either the user or the whole server from doing those actions. For example: A user promotes 5 users within x mins, then that user is blocked for xx minutes from promoting anyone, or the server is blocked from promoting anyone. This is based on config elements, so multiple rules can be set. 

As you can see, it's huge.
This is still W.I.P (Nothing has been done yet), but it's a good preview of what's coming. 
Kind regards from me :)


Kevinking500:
Just added the anti-raid config to the changelog, it just came to my mind randomly

PR #208, opened 2026-08-06
Welcomer: Fixed issue with duplicated message parameter

Timon611:
Hey, regarding to the support ticket (ID: 1534784220763914320) I removed the duplicated parameter (since the %mention% parameter is used for every case, even for unboosting) 🙂


Timon611:
Hey @SCDerox, I would create a bug report for the user once, so he/she gets notified for any updates on this, enjoy your vacation 🏝️

SCDerox (approved):

PR #209, opened 2026-09-06
Ping Protection V2.5

Kevinking500:
# Ping protection V2.5 - next-gen ping moderation and QoL improvements

**Changelog:**
- **Moderation rules modularity:** Moderation rules now have an option to let it apply to select type of pings (mention, reply or both)
- **Ping type clarified:** Ping types are now clarified in the user's history, showing if it was a reply ping or a mention.
- **Improved moderation rule timeframe:** The timeframe of each moderation action is no longer limited to days - users can now input, for example, 10s and it will check for the actions done in the last 10s for that moderation rule. When no text is added, it falls back to days (so old configs will stay to their days configuration to avoid breaking them)
- **Time-framed data deletion:** Higher ups can now decide to only delete data from a later date (so for example deleting data that is >7d old instead of everything)
- **Automatic warning message deletion:** Warning messages clutter up channels. Users can now choose if and after how long the message auto-deletes.
- **Toggle your protection via a command:** Some staff don't mind being pinged while being active in the server, so a new subcommand now allows protected users to toggle their protection status. If they don't mind being pinged for some time, they toggle it off, and back on after some time. The protection toggles back on automatically after 24h to prevent staff turning their protection off while they should be protected from the config.
- **Longer leaver data retention:** Leaver data retention can now be kept for up to 60d instead of 7d, handy for investigations and similar, idk why I kept the limit to 7d before. Default is now also 7d.
- **Kick DM message:** Users are now DMed a customizable message when they are kicked because of a moderation action instead of being kicked into the void
- And some other QoL improvements, like shortened confirmation phrase to prevent translation issues
- *Possible extra feature: I might decide to also support voice channels in whitelisted channels, haven't even tested vc's yet, ever*

AI usage:
Some thinking to think of some features. Also writing the migrations file.
This will likely be the last update of this module as you know my priorities shifted, though unless crucial (big bug reports etc) things appear.

### DIFF
--- a/modules/ping-protection/models/UserPingPreference.js
+++ b/modules/ping-protection/models/UserPingPreference.js
@@ -1,0 +1,30 @@
+const {
+    DataTypes,
+    Model
+} = require('sequelize');
+
+module.exports = class PingProtectionUserPingPreference extends Model {
+    static init(sequelize) {
+        return super.init({
+            userId: {
+                type: DataTypes.STRING,
+                allowNull: false,
+                unique: true
+            },
+            disabledUntil: {
+                type: DataTypes.DATE,
+                allowNull: true,
+                defaultValue: null
+            }
+        }, {
+            tableName: 'ping_protection_user_preferences',
+            timestamps: true,
+            sequelize
+        });
+    }
+};
+
+module.exports.config = {
+    name: 'UserPingPreference',
+    module: 'ping-protection'
+};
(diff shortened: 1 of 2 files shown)


Kevinking500:
ETA: Tomorrow. Config and locales sided things have been done, only some logic things have to be done & some testing then

Kevinking500:
Lovely failing tests.. Gtg to school now so i'll see later

Kevinking500:
Holy moly thank Copilot for this ragebaiting of a fix (i'll probs have to do it again later, idk why I focussed on doing ts now)

